PR #22, opened 2022-03-14
Attempted some fixes & optimizations

unquietwiki:
Hey there. Been using this mod for a while now, and wanted to take a stab at resolving some potential bugs, and seeing if I could speed things up a little. I hope this is of use to you. Thanks for maintaining this mod; it helps keep the game running smooth!

### DIFF
--- a/GameData/KSPCommunityFixes/KSPCommunityFixes.version
+++ b/GameData/KSPCommunityFixes/KSPCommunityFixes.version
@@ -2,7 +2,7 @@
   "NAME": "KSPCommunityFixes",
   "URL": "https://raw.githubusercontent.com/KSPModdingLibs/KSPCommunityFixes/master/GameData/KSPCommunityFixes/KSPCommunityFixes.version",
   "DOWNLOAD": "https://github.com/KSPModdingLibs/KSPCommunityFixes/releases",
-  "VERSION": {"MAJOR": 1, "MINOR": 7, "PATCH": 0, "BUILD": 0},
+  "VERSION": {"MAJOR": 1, "MINOR": 8, "PATCH": 0, "BUILD": 0},
   "KSP_VERSION": {"MAJOR": 1, "MINOR": 12, "PATCH": 3},
   "KSP_VERSION_MIN": {"MAJOR": 1, "MINOR": 8, "PATCH": 0},
   "KSP_VERSION_MAX": {"MAJOR": 1, "MINOR": 12, "PATCH": 3}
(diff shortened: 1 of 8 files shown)


unquietwiki:
@gotmachine Understood, and due apologies for the headaches I apparently gave you on this. What I do have running now is not giving any immediate errors, and I think my RAM usage dropped quite a bit too.  There were two goals in this that I wanted to accomplish...

1. I was having some really laggy ship builds in VAB with not even 50-100 parts. I know some of the fixes here are meant to deal with VAB errors and issues; I was hoping the forced LINQ Lists would speed that up.
2. https://github.com/microsoft/Microsoft.Unity.Analyzers/blob/main/doc/UNT0008.md → this is the error that was being flagged, that's come up in some other mods & whatnot; this is what I tried to fix, and if you're confident it's not an issue, then it's not an issue. [This thing I wrote up](https://unquietwiki.com/en/TechFindings/UnityNulls) are my notes on this particular conundrum; I wasn't even aware it was a potential issue until another mod dev brought it to my attention.

unquietwiki:
@gotmachine in replying to your other concerns, I noticed you mentioning https://github.com/KSPModdingLibs/KSPCommunityFixes/blob/master/KSPDevPath.xml.copyMe ; which I didn't realize was in place; due apologies on that. That's actually a pretty neat solution to what I've seen in largely inconsistent behavior regarding references in other mods. Is this something that others should be using?

gotmachine:
> I was having some really laggy ship builds in VAB with not even 50-100 parts

It's quite unlikely anything in this mod can affect performance or memory usage negatively in any measurable way. But if you can consistently measure a performance difference with and without the mod installed, there is likely a bug lurking around somewhere and in that case, please open an issue and try to provide additional information.

> I was hoping the forced LINQ Lists would speed that up.

That can only have the exact opposite effect, both on speed and memory usage.

> That's actually a pretty neat solution to what I've seen in largely inconsistent behavior regarding references in other mods. Is this something that others should be using?

That's my custom solution, which has some downsides too, and it's not the "right way" either.
In a "regular" csproj with hardcoded references, you can usually point to your local KSP `KSP_x64_Data\Managed` folder in the project properties > reference path menu of VS, which will create a local git-ignored `*.csproj.user` file containing that reference path. This should allow visual studio (or whatever IDE you are using) to find the *.dll it needs without having to modify the original `*.csproj`. More detailed explanation : https://lowleveldesign.org/2011/04/20/take-advantage-of-referencepath-in-visual-studio-and-debug-locally-external-libraries/

unquietwiki:
@gotmachine I took your guidance & tried a hopefully better set of things for you to look at. I left the Unity errors in place as warnings, so you can see what I was seeing & make any desired adjustments. VAB & memory performance seems good at the moment.

https://github.com/unquietwiki/KSPCommunityFixes

PR #29, opened 2022-04-07
New QoL Feature: allow hide extended info of PartUpgrade parttooltips

NathanKell:
Part Upgrades hijack the existing part tooltip system, but (either because I was dumb at the time, or because it atrophied later, can't remember at this point) when you right-click to get them to stick (and so you can purchase them), the extended info tab of the base AvailablePart/PartPrefab used for the icon gets displayed. Ugh.

This simple fix will prevent it from displaying.


NathanKell:
Defaulted to enabled per discussion with @gotmachine

PR #31, opened 2022-04-19
Hide delta V UI when the stock delta V app/calculations are disabled.

NathanKell:
Otherwise it's just showing a bunch of "0m/s" and empty UI windows, which looks broken.

Annoyingly there's a little empty gray area where the total dV display was on the staging UI, but I'm not seeing an obvious way to remove it or shrink the background, so eh. Better than a constant display of "0m/s" there at least.

### DIFF
--- a/GameData/KSPCommunityFixes/Settings.cfg
+++ b/GameData/KSPCommunityFixes/Settings.cfg
@@ -74,6 +74,12 @@ KSP_COMMUNITY_FIXES
   // Fix deployable parts (antennas, solar panels, radiators...) always starting in the extended 
   // state when the model isn't exported in the retracted state. Affect parts from various mods.
   ExtendedDeployableParts = true
+  
+  // Fix the delta V data (and info panels) on stage UI elements showing
+  // when the delta V app and delta V calculations are disabled.
+  // Also hide extended burn time info on the navball in this case, because
+  // it depends on the stock delta V implementation.
+  DeltaVHideWhenDisabled = true
 
   // ##########################
   // Obsolete bugfixes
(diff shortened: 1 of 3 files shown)


gotmachine:
Seems good.

A few remarks :
- You don't need prefix methods to return a bool if they never have to skip the original, you can just just have a void return type method.
- You can use use a `out bool __state` argument in the prefix and `bool __state` argument in the postfix for situations where you want to pass some info between them, this is a bit safer than using a static variable (see https://harmony.pardeike.net/articles/patching-prefix.html#passing-state-between-prefix-and-postfix).

I was about to do a new release, will merge and take care of those bits. Thanks for the contribution !

NathanKell:
Ooh, thanks! Will remember. :)

gotmachine:
Little note in case the become relevant in the future : the patch will break the [BasicDeltaV](https://github.com/DMagic1/KSP_BasicDeltaV) mod by DMagic. 

Doesn't really matter since that mod is abandoned and doesn't work in 1.12, but in case it ever get revived, it should disable the patch through MM.

yalov:
this "burn time" string is also could be removed for disabled stock dV


and maybe put "disable Stock dV calculation" in the KSPCF GUI options?

gotmachine:
> this "burn time" string could be removed for disabled stock dV

Yep I noticed that too. Will try to look into it.

> maybe put "disable Stock dV calculation" in the KSPCF GUI options?

Well, I prefer not to make that an user-facing setting. I remember seeing some plugins relying on those (BetterBurnTime ?). "Advanced" users can take the responsibility of editing the KSP `settings.cfg` file manually to disable them.

This patch is there as a convenience fix for disabling the UI for that use case or when another plugin (RO...) decides to disable the stock DV calcs. Resulting bugs or plugin incompatibilities while doing so are their responsibility.

NathanKell:
Oh yuck sorry I missed the text. I can also look into it this evening,
think it's just 'ebtText'

Sent by my thumbs, slowly.

PR #36, opened 2022-05-09
For archived contracts, show accepted/finished dates.

NathanKell:
Small QoL feature: right now KSP shows accepted/offered/deadline/etc dates for offered and active contracts, but shows *no date at all* for archived contracts. This patch makes it so that archived contracts show the date they were accepted and the date they completed/failed/etc.

Defaults to false, which I think is correct? I mean, *I* would always play with it turned on, but I'm not 100% comfortable making that call be the default.

### DIFF
--- a/GameData/KSPCommunityFixes/Settings.cfg
+++ b/GameData/KSPCommunityFixes/Settings.cfg
@@ -122,6 +122,9 @@ KSP_COMMUNITY_FIXES
 
   // Append "[Auto-Saved Craft]" when relevant to the craft name in the Launchpad / Runway UI
   AutoSavedCraftNameAtLaunch = true
+  
+  // Show date a contract finished when displaying info on a finished contract in Mission Control
+  ShowContractFinishDates = false
 
   // ##########################
   // Performance tweaks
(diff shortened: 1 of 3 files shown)


NathanKell:
And once again a transpiler might be nice but it didn't feel worth it. Truly minor about of GC and it's in UI code anyway.

gotmachine:
> And once again a transpiler might be nice but it didn't feel worth it. Truly minor amount of allocs and it's in UI code anyway.

Yeah, given that the whole thing uses string concatenation all over the place instead of StringBuilders this isn't gonna make any significant difference in terms of GC allocs.

> Defaults to false, which I think is correct? I mean, I would always play with it turned on, but I'm not 100% comfortable making that call be the default.

Feels like worst case scenario would be the patch doing nothing, so probably fine to enable it by default ?
Did you test this in conjunction with ContractConfigurator ? It has a `MissionControlTextRich()` override, but looking at it, should be fine : 
https://github.com/jrossignol/ContractConfigurator/blob/7015264e285f00d4d2772888d8b9e1d992b34fc1/source/ContractConfigurator/ConfiguredContract.cs#L618-L630

NathanKell:
Worst case scenario is the text looks wrong in other languages, but if it does then the stock date text for offered contracts probably does too. Other than that, yeah, worst is it does nothing.

Hilariously I didn't realize CC did that, so...I've *only* tested it with CC.

NathanKell:
(By looks wrong I mean `translated word + ":" + date` is what this outputs, but so does stock code.)

PR #44, opened 2022-06-19
Fix bug where vessels can remain in situation Splashed.

NathanKell:
This does two things:
* A vessel that spawns splashed will stay in prelaunch until it is moving at > 1m/s (like how a landed vessel will stay in prelaunch until it is moving >2.5m/s). Normally splashed clobbers prelaunch, which is Bad, especiallby because:
* Fix bug where  a vessel can never go unsplashed if it loses its splashed parts. The only way for something to lose splashedness in stock is for all splashed parts to lose contact with water, but if the vessel loses (decouples or they die) splashed parts, its splashed bool is never reset (the splashed check, like the landed check, is event driven rather than polled).

Note this all is slightly confusing because there is both a vessel state (SPLASHED, PRELAUNCH, FLYING, etc) and the bool vessel.Splashed (like vessel.Landed) which says whether any parts are splashed.


NathanKell:
This has now been tested in the RO discord by users who did have the bug (and by me, though I could never repro the bug >.> but at least it doesn't hurt anything for me) and seems to work fine.

PR #47, opened 2022-06-22
Fix not applying DepartmentConfig.HeadImage

NathanKell:
Jim apparently forgot to hook up HeadImage when loading the Department list in the Admin UI. This fixes that.

(Strategia has a clobber-based version of this fix, but it only sometimes works in my experience. It won't *stop* working with this fix, so I have no concerns with breaking compat.)


NathanKell:
@gotmachine it's...plausible this should be a Modding support thing rather than a bugfix. Depends on your point of view I guess. I'm fine either way, obviously.

PR #53, opened 2022-07-04
Allow app launcher to collapse/expand (even in SC UIs)

NathanKell:
Remaining wrinkle: the white bits of the Contracts and Messages buttons don't show when expanded in MC/AC/etc, because of how the rest of the applauncher has to have its canvas sorting order changed to render in those UIs. I'm not quite familiar enough with Unity UI to know if adding another Canvas to their buttons and leaving its override off except then would fix it, or if *every* child would need its canvas managed...


gotmachine:
The issue with some button icons not showing is because they are using a sprite renderer (the icon is animated), and those are independent from the canvas in terms of sorting layer / order. Will try to fix it.

PR #57, opened 2022-07-12
Fix #55: Add ProgressTrackingSpeedBoost

### LINKED ISSUE #55: Performance issue: ProgressTracking.Update seems to be O(bodies * vessels)
In my "twitch plays KSP" stream I have a game save with galaxies unbound + lots of other planet packs (320 bodies in total) and LOTS of kerbals.  It was running slow so I decided to attach a profiler and was astonished to see this: 
image

A brand new save in the same universe has ProgressTracking.Update at about 5ms, which is still astonishingly high but is likely related to the number of bodies.

I haven't looked into what this function is doing, but it would be awesome to optimize this.

JonnyOThan:
KSP's stock ProgressTracking evaluates every progress node for every vessel in the universe.
There are ~19 progress nodes per celestial body, and this can get extremely expensive for mods with a
large number of bodies or saves with a large number of vessels.
This tweak only evaluates the currently active vessel and removes the per-body checks which don't
have any effect, at least in the stock game.

### DIFF
--- a/GameData/KSPCommunityFixes/Settings.cfg
+++ b/GameData/KSPCommunityFixes/Settings.cfg
@@ -206,6 +206,13 @@ KSP_COMMUNITY_FIXES
     LogGameEventsSubscribers = false
   }
 
+  // KSP's stock ProgressTracking evaluates every progress node for every vessel in the universe.
+  // There are ~19 progress nodes per celestial body, and this can get extremely expensive for mods with a
+  // large number of bodies or saves with a large number of vessels.
+  // This tweak only evaluates the currently active vessel and removes the per-body checks which don't
+  // have any effect, at least in the stock game.
+  ProgressTrackingSpeedBoost = true
+
   // ##########################
   // Modding
   // ##########################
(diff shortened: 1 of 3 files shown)


gotmachine:
Evaluating only the active vessel would be a breaking change, not only for "world firsts", but also for contracts.

The part about removing the per body update delegate feels alright. Not 100% safe, but I doubt anybody ever added custom tracking nodes.

JonnyOThan:
I’ll do a more diligent search…I’m pretty sure all the contract related stuff is not things that are handled by OnVesselIterate

JonnyOThan:
Confirmed: Every stock class that populates the OnIterateVessels callback only looks at the active vessel inside the callback.

gotmachine:
Ok, did a quick review of the ContractConfigurator codebase too, couldn't find anything that might be affected.
Also found a bunch of other mods manipulating `ProgressTracking` stuff in a way or another, but it doesn't seem any of them is relying on `IterateVessels` callbacks.

Will merge for next release, thanks for the contribution.

PR #58, opened 2022-07-13
Fix #56: Fix CommnetNetwork.Update performance

### LINKED ISSUE #56: Performance issue: CommnetNetwork.Update is slow
I have a save with a lot of vessels and bodies in flight, and this function is roughly 11% of the frame time:
image

Is there anything that can be done to optimize it?

JonnyOThan:
The throttling logic that was supposed to update it once every 5 seconds was broken

### DIFF
--- a/GameData/KSPCommunityFixes/Settings.cfg
+++ b/GameData/KSPCommunityFixes/Settings.cfg
@@ -224,6 +224,18 @@ KSP_COMMUNITY_FIXES
   // have any effect, at least in the stock game.
   ProgressTrackingSpeedBoost = true
 
+  // Enables the commnet performance optimizations
+  // The stock game is supposed to throttle commnet updates to once every 5 seconds when the active vessel
+  // is unpacked, but there is a bug that prevents this from working properly.  Enabling this setting
+  // fixes the bug, and you can optionally customize the timing intervals with the settings below.
+  CommNetSpeedBoost = true
+
+  COMM_NET_PERFORMANCE_SETTINGS
+  {
+    packedInterval = 0.5
+    unpackedInterval = 5
+  }
+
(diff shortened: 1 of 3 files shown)


gotmachine:
For reference, the original logic is :
```cs
if (!queueRebuild && !commNet.IsDirty)
{
    double currentTime = Time.timeSinceLevelLoad;

    double activeVesselInterval = 
        (FlightGlobals.ActiveVessel == null || FlightGlobals.ActiveVessel.packed) 
            ? packedInterval 
            : unpackedInterval;

    if (prevUpdate + activeVesselInterval > currentTime)
    {
        double focusedVesselInterval = focusedVessel == null ? 0.0 : packedInterval;
        if (prevUpdate + focusedVesselInterval > currentTime)
        {
            graphDirty = true;
            return;
        }
    }
}
```
You patch is entirely bypassing the first two checks (`if (!queueRebuild && !commNet.IsDirty)`), that doesn't seem right to me.

I think the rest kinda works, although rate-limiting based on real-time instead of in-game time is quite questionable, but I guess that's beyond the point.

JonnyOThan:
I can add those checks back in, but those two fields are never set to true in the stock code.

gotmachine:
I will make the patch disabled by default. 

Even if the throttling mechanism is originally intended, this patch introduce a breaking change from CommNet usually being in sync with the game loop to it using potentially vastly outdated data. 

0.5 RL seconds updates at 100000x warp means essentially garbage results, and while this doesn't matter too much in stock, there are mods relying on CommNet for various stuff.

JonnyOThan:
Thanks!  But note the stock code ALSO can do 0.5 second intervals, if focusedVessel is not null.

PR #94, opened 2022-09-13
Add RestoreMaxPhysicsDT

JonnyOThan:
After using physics warp, Unity's max physics dt will never be returned to the value specified in game settings which can degrade performance in some cases.

Reported by Wilds on KSP Forum: https://forum.kerbalspaceprogram.com/index.php?/topic/184740-solved-losing-300-performance-after-time-warping/


JonnyOThan:
hm I'm not sure why this is conflicting!  Let me redo it..

PR #95, opened 2022-09-13
Add RestoreMaxPhysicsDT

JonnyOThan:
When using physics warp, Unity will set the max physics dt to be at least as high as the scaled physics dt.  But KSP will never restore it back to the normal value from the settings.  This can degrade performance as it allows more FixedUpdates to run per frame.

reported by Wilds on the forum: https://forum.kerbalspaceprogram.com/index.php?/topic/184740-solved-losing-300-performance-after-time-warping/

### DIFF
--- a/GameData/KSPCommunityFixes/Settings.cfg
+++ b/GameData/KSPCommunityFixes/Settings.cfg
@@ -292,6 +292,10 @@ KSP_COMMUNITY_FIXES
   // speed up reading) at the cost of human readability.
   SkipIndentsOnSavesAndCraftFiles = false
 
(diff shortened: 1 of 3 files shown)


NathanKell:
What min version should this have? My guess is it's safe for 1.8+ but there's no min specified so it'll default to 1.12 min.

JonnyOThan:
Oh good question. According to that forum post it’s existed since at least 1.6, and since this is a prefix it should work as long as the target method exists.  I’d guess that has been there since the dawn of time (heh).  But it should probably be tested and verified before marking it safe for earlier versions.

PR #120, opened 2023-02-05
Fix MapSOCorrectWrapping removing stock MoHole

R-T-B:
Just blacklists the fix from a body named Moho.  Thought it might be handy.

### DIFF
--- a/GameData/KSPCommunityFixes/Settings.cfg
+++ b/GameData/KSPCommunityFixes/Settings.cfg
@@ -154,6 +154,10 @@ KSP_COMMUNITY_FIXES
   // Avoids big spikes on the poles and incorrect biomes where the oles have disparate biomes.
   MapSOCorrectWrapping = true
 
+  // Fix biome and heightmap texture wrapping from pole to pole, even on Moho
+  // This may wipe out the Mohole which is based on said bug.
+  MapSOCorrectWrappingPlusMoho = false
+
   // Fix spread angle still being applied after decoupling symmetry-placed parachutes.
   ChutePhantomSymmetry = true
 
--- a/KSPCommunityFixes/BugFixes/MapSOCorrectWrapping.cs
+++ b/KSPCommunityFixes/BugFixes/MapSOCorrectWrapping.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using HarmonyLib;
 using UnityEngine;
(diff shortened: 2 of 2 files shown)


gotmachine:
Mmmh... This is quite fragile...

First, this shouldn't be checking the config on every call. `ConstructBilinearCoords()` is a performance critical method, this add a significant overhead, easily avoidable by putting the parsed state in a static field.

But main issue is that there is no guarantee that in the call stack leading to a `MapSO.ConstructBilinearCoords()` call, the `MapSO` instance is matching `FlightGlobals.currentMainBody`. There are multiple stock API methods where this can happen (and I'm pretty sure they are used in such a way by some mods), and if remember right even PQS generation isn't always happening on the current main body either.

`MapSO` instances should be kinda static once PSystemSetup is done (even with Kopernicus I think ?), so it should be possible to get static references to the ones matching stock Moho and do direct reference comparisons. `MapSO` has a `_name` field and derive from `ScriptableObject`, which also has a `name` property, both are populated (at least for stock bodies), but this don't feel like a 100% reliable check either. A safer way to ensure we're targeting the right `MapSO` instances would be to compute a strong hash from the `byte[] _data` field, and store the hashes for the stock Moho `MapSO` instances as constants to compare what we get at runtime.

I will investigate how to do this :)

R-T-B:
Interesting.  I had no idea it fired that often.  Anyhow, I leave this idea in your evidently capable hands.

gotmachine:
> One thing to keep in mind about data hashes vs simple name comparisons though: we have no idea if that will work with say, a rescaled moho, do we?

`MapSo` is essentially the raw texture data. I don't know how rescaling bodies work but I doubt it's by rescaling the texture itself.
Anyway, I won't do a real hash check, that would add too much overhead to loading, I will just check the name and a few bytes in the array.

And on a side note, I don't really like that we are doing this patch at all. Like some other cases we discussed, this is a Kopernicus specific issue, and it should be handled by Kopernicus, not by KSPCF. I just did some limited testing, and this patch in fact significantly alter the poles terrain on all stock bodies, no only Moho. 

I don't argue that the stock code has a bug (which Squad decided to turn into a feature), but these kind of issues are exactly why I started KSPCF with the mission statement that its changes should be as seamless as possible in stock, and to let modded KSP issues be handled by other mods.

gotmachine:
Closing this, see https://github.com/KSPModdingLibs/KSPCommunityFixes/issues/121

R-T-B:
Closing as you did your own fix.

PR #130, opened 2023-03-26
SunFix

IIodyne:
Fixes gravitational properties of Sun. See comments in file.


gotmachine:
I was about to say "nice find", but after checking by using in-game code directly, the reason you're finding the discrepancy in the derived gravity constant for the sun is because your input data for the sun mass is wrong.

You are using `1.75656696858329E+28` for the sun mass, it is actually `1.75654591319326E+28`. With that value, the resulting  gravity constant is consistent with the `6.67408E-11` value found by using the parameters of the other stock bodies (actually to be 100% accurate, it's `6.6740800000000003E-11`).

image

I used the following code to get the input data :
```cs
foreach (CelestialBody body in FlightGlobals.Bodies)
    bodyInfo += $"\n{body.name};{body.Mass:G17};{body.gravParameter:G17};{(body.gravParameter / body.Mass):G17}";
```

I'm not sure what was the source of your data, but that source is incorrect. Aside, doing a quick review of what the modding ecosystem has to say, I've found this issue : https://github.com/KSP-KOS/KOS/issues/2409, which highly suggest that things have been that way since at least KSP 1.4.5.

Additionally, the KSP official changelog (ie, readme.txt) has this mention in the KSP 1.2.0 release notes : 
> Now always use g0 = 9.80665 and G (big G) = 6.67408e-11 for gravitational constants.

My guess is that your source is using data from a very old KSP version where this issue was actually present.

IIodyne:
Ah, you are correct; thanks for verifying! You are indeed right I used the kittopia dumps. I did have the thought that they were quite old but I could not find any update on it unlike you were able to. Are there updated values akin to those kittopia dumps I can access (for other uses)?

gotmachine:
Well, the latest commit of the kittopia dumps are from KSP 1.9.1 : 
https://github.com/Kopernicus/kittopia-dumps/blob/62d64791adba14aacd33b5f2d7a3c42f8b78cd8b/Configs/Sun.cfg#L19
I doubt there has been any significant changes since then, but you could try to open an issue on their repo asking for an update.

PR #140, opened 2023-04-12
Fix landing gear light patch so it applies to both parts

JonnyOThan:


gotmachine:
Whoops. Thanks, good catch.

PR #151, opened 2023-06-05
FastBiomeQuery

gotmachine:
The stock biome maps (`CBAttributeMapSO` class) store a texture as RGB color data in a byte array where every pixel takes 3 bytes. Each biome definition (`CBAttributeMapSO.MapAttribute`) is given a specific color, and is stored in an array.
Querying a biome at a position (`CBAttributeMapSO.GetAtt()`) is a complex method that perform bilinear interpolation on that color array, attempts to find the closest color in the biome definition array, and does a bunch of heuristics in case the texture colors don't exactly match the ones defined in the biome definitions.

Some issues with this :
- Biome querying is quite slow as a result, which can be a bottleneck with mods doing a lot of such queries (Kerbalism, ScanSAT, Parallax scatters, ContractConfigurator...) , but with some stock code too (Contracts, KerbNet...).
- Storing full pixel color data takes up a lot of unnecessary RAM
- The stock bilinear interpolation has a bug resulting in jagged transitions between biomes along a specific direction

This is a replacement for the stock class, implemented as a `KSPCFFastBiomeMap` class deriving from `CBAttributeMapSO` :
- Instead of storing color data taking 3 bytes per pixel, store biome indices, taking 1 byte per pixel, reducing memory usage by a factor 3.
- Ability to reinterpret the pixel color data as biome indices (either from an already compiled `CBAttributeMapSO` or from a `Texture2D`), following the same heuristics as the stock `GetAtt()` method, allowing upfront validation of biome map, and preventing having to do it on every query.
- Reimplement the `CBAttributeMapSO.GetAtt()` method doing a direct lookup for the biome index instead of parsing colors, performing a correct bilinear interpolation.
- Fix incorrect coordinate wrapping at the north pole (stock bug), resulting in the south pole biome being returned near the north pole.

The benefits are :
- Reduced memory usage : 
  - Stock : from 189MB to 63MB
  - RSS : from 82MB to 27MB
  - OPM : from 549MB to 183MB
- An order of magnitude faster, and constant time biome querying (`GetAtt()`), unaffected by biome count. On average 15-20x faster than the stock implementation, up to 30x on bodies with a large amount of biomes.

Drawbacks :
- This introduce a limitation in that amount of biomes per body is limited to 256 whereas the stock implementation supports a virtually infinite amount (2^24). I doubt this can be an issue in practice.
- This require an upfront conversion of all the `CBAttributeMapSO` instances into a `KSPCFFastBiomeMap`. This is performed after `PSystemSetup` to catch Kopernicus changes, and before the main menu is loaded. The conversion is multithreaded, but it will introduce a perceptible delay on lower end machines. For reference, on a 5800X3D : 
  - Stock maps conversion, 1 thread : 1.4s
  - Stock maps conversion, 16 threads :  0.2s
  - Stock + OPM maps conversion, 1 thread : 3.2s
  - Stock + OPM maps conversion, 16 threads : 0.5s

While I think this is acceptable, it should be noted that doing this in the context of biome maps loaded by Kopernicus is utterly inefficient, as the complete chain of events is something like : 
- Kopernicus load the biome texture files as a `byte[]` from disk
- For some texture formats, it first parse it as a `Color32[]`, then copy it as a `Texture2D`, in other cases it relies on Unity to perform the conversion from the `byte[]` to a `Texture2D`.
- It feed that texture to the stock `CBAttributeMapSO` parser, which : 
  - Convert it as a `Color32[]`
  - Convert the `Color32[]`  to a 3Bpp RGB data `byte[]`
- Then KSPCF uses that RGB 3 Bpp `byte[]` to produce a biome indices 1Bpp `byte[]`, with a `Color32[]` conversion in between.

It would be vastly more efficient for Kopernicus to directly feed the KSPCF implementation the `Texture2D` or `Color32[]` without the useless conversion to a `CBAttributeMapSO` in between. This is difficult to implement from the KSPCF side (at least efficiently). Aside, Kopernicus would benefit from an overhaul of its biome querying stuff, so it seems more sensible to actually have a side by side implementation of the KSPCF fast biome in Kopernicus.

From a practial PoV, I will be AWOL from the KSP modding scene for some time, effective right now. Consequently, I will delay pushing this patch to KSPCF in case there are issues popping up that require some fixing. Plus I would like to contribute that work to Kopernicus, ideally synchronizing that feature release both in KSPCF and Kopernicus.

In the meantime, here is a build with the changes, for those willing to test it : 
[KSPCommunityFixes_1.29.0_FastBiomeQueryPR.zip](https://github.com/KSPModdingLibs/KSPCommunityFixes/files/11651341/KSPCommunityFixes_1.29.0_FastBiomeQueryPR.zip)

For reference, this illustrate the bug with biome transitions in the stock bilinear sampler :
image

### DIFF
--- a/GameData/KSPCommunityFixes/Settings.cfg
+++ b/GameData/KSPCommunityFixes/Settings.cfg
@@ -154,7 +154,7 @@ KSP_COMMUNITY_FIXES
   // Avoids big spikes on the poles and incorrect biomes where the oles have disparate biomes.
   MapSOCorrectWrapping = true
   // The Mohole is a result of the bug the MapSOCorrectWrapping patch is fixing. If set to
-  // true, the patch won't apply to the stock Moho height/biome maps
+  // true, the patch won't apply to the stock Moho height map
   MapSOCorrectWrappingIgnoreMoho = true
 
   // Fix spread angle still being applied after decoupling symmetry-placed parachutes.
--- a/KSPCommunityFixes/BugFixes/MapSOCorrectWrapping.cs
+++ b/KSPCommunityFixes/BugFixes/MapSOCorrectWrapping.cs
@@ -1,6 +1,7 @@
 ﻿using HarmonyLib;
 using System;
 using System.Collections.Generic;
+using System.Runtime.CompilerServices;
 using UnityEngine;
 
 namespace KSPCommunityFixes
@@ -26,8 +27,8 @@ namespace KSPCommunityFixes
             // see https://github.com/KSPModdingLibs/KSPCommunityFixes/issues/121
             // The patched ConstructBilinearCoords() methods will have the side effect of removing the Mohole because Squad
             // choose that "this is not a bug, it's a feature".
-            // To prevent it, we acquire references to the Moho MapSO instances (biome map and height map) and fall back
-            // to the original stock implementation if the method is called on those instances.
+            // To prevent it, we acquire references to the Moho MapSO heightmap instance and fall back to the original stock
+            // implementation if the method is called on those instances.
             // Note that all stock bodies will actually get a significantly different terrain at the poles due to this patch,
             // but due to how performance sensitive those method are, checking all stock MapSO instances (there are 40 of them)
             // isn't really an option.
@@ -46,8 +47,9 @@ namespace KSPCommunityFixes
             }
         }
 
-        private static MapSO moho_biomes;
         private static MapSO moho_height;
+        private static readonly double lessThanOneDouble = StaticHelpers.BitDecrement(1.0);
+        private static readonly float lessThanOneFloat = StaticHelpers.BitDecrement(1f);
 
         static void PSystemSetup_Awake_Postfix(PSystemSetup __instance)
         {
@@ -58,15 +60,7 @@ namespace KSPCommunityFixes
 
             foreach (MapSO mapSo in so)
             {
-                if (mapSo.MapName == "moho_biomes"
-                    && mapSo.Size == 6291456
-                    && mapSo._data[0] == 216
-                    && mapSo._data[1] == 178
-                    && mapSo._data[2] == 144)
-                {
-                    moho_biomes = mapSo;
-                }
-                else if (mapSo.MapName == "moho_height"
+                if (mapSo.MapName == "moho_height"
                          && mapSo.Size == 2097152
                          && mapSo._data[1509101] == 146
                          && mapSo._data[1709108] == 162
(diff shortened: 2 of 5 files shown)


R-T-B:
I finally got a chance to test this...

As it were, this has major issues with mods that use nonprecise biome colors, usually due to compression artifiacts.  Parallax is one such mod as it does its Kerbin biome map as a .dds for some inconceivable reason.  The fix is easy though, remake it as a png.  I have attached my attempt at that below:

biomeMap

Then just point the config at the new biome map.  This could easily be made a PR in parallax, but for other less significant mods, concerns remain about how compression artifacts will be handled.  The results in my testing were wildly off from what they should have been (thought I was in poles at the launchpad).

Can this be improved?  I don't really know, but that's the report.  Otherwise works well.

gotmachine:
> The results in my testing were wildly off from what they should have been (thought I was in poles at the launchpad).

That's not intended. Results can differ slightly from stock, especially at biomes transitions but they shouldn't be "wildy off", even with a biome map using non-configured colors.
Can you give the exact mod/body involved and lat/lon coordinates of problematic places ?

R-T-B:
It should happen with Kerbin launchpad with stock paralax using its included biome map.  I did have it rescaled to about 5x.  I doubt that makes a difference though.  Will run some additional tests.

R-T-B:
In further testing, this is still going on, ie not biome map related.

I am unsure why exactly, but with this changeset, my KSP launchpad always identifies as "Southern Ice Shelf" biome.  I will try to isolate what mods exactly lead to this conclusion, but one thing is certain:  Turning off this changeset and going back to the old biome code remove the abnormality.

R-T-B:
I finally sorted this out (by getting farther than stock Kerbin in my save lol).

The problem is Sigma Dimensions.  Rescaling stock worlds does not work with this changeset, period.  Disabling BiomeMapOptimizer appears to fix it, but I have a feeling that's really reducing the utility of this, no?

gotmachine:
I'm a bit confused...
Can you give a clear minimal repro setup/install ?

gotmachine:
On a side note, I did a bit of perf profiling between the Kopernicus "biome cache" used by Parallax and the FastBiomeQuery
implementation.
FastBiomeQuery is consistently 5-10x faster than the Kopernicus biome cache when it is at its default level (`ScatterLatLongDecimalPrecision = 5`), and that's when the cache is already populated. 
When it's not, the cost is closer to 20-50x, and this is the case in many situations (for example low/mid altitude flying in a straight line). Looking at the cache implementation, there are a few inefficiencies, but I don't think that would change the overall figure, the dictionaries lookups are more costly than just querying the biome directly.

I think I will look into pushing the FastBiomeQuery implementation as a Koperncius PR as well as a KSPCF patch. The existing Kop biome sampler can be rewired to use it, and I can disable the KSPCF patch when Koperncius is present.

R-T-B:
> @R-T-B I'm a bit confused... Can you give a clear minimal repro setup/install ? I don't see any issue when using sigma dimensions 0.10.8 + kopernicus 190, with the following sigma config :
> 
> ```
> 	Resize = 5
> 	Rescale = 5
> ```
> 
> Kerbin biomes are showing as expected in the F12 menu map view biome overlay and by checking in flight in the "flight info" tab at various locations.
> 
> I also tried throwing parallax into the mix and I don't see any obvious issue with scatter placement

that's odd.  Let me do some more testing today to try to better isolate my issue, and get back to you.

R-T-B:
A quick thought, as it may be a day or two before I can get a test environment up and running:

The buggy behavior was noticed when in flight, not landed, so unsure if that makes a difference with how you are testing.

gotmachine:
> it came up as faulty logged kerbalism experiments

Ha, well, good catch, but I'm pretty sure the issue is with Kerbalism. I don't remember what I did exactly, but I remember rolling out a hack to get biomes faster in there too. Will look into fixing it from the Kerbalism side (as whatever I did won't be relevant anymore following the release of this).

R-T-B:
Sounds good.  The latest changes on Kopernicus were a simple single-file hotfix, it should be easy to rebase.

gotmachine:
For reference, the [Kopernicus implementation of this patch](https://github.com/Kopernicus/Kopernicus/pull/601) has been released.
Since the KSPCF patch only tries to convert the stock `CBAttributeMapSO` type in the [`OnDestroy()` of `PSystemSpawn`](https://github.com/KSPModdingLibs/KSPCommunityFixes/blob/ffa2b9ce782a5dd6e2f40811f08271fb5fa88779/KSPCommunityFixes/Performance/FastBiomeQuery.cs#L594C14-L594C25), maps already converted by Kopernicus in `PSystemSpawn` are of type `KopernicusCBAttributeMapSO` and should be ignored by KSPCF, so the two implementations should be able to coexist without issues.

PR #160, opened 2023-09-03
Make docking ports conserve momentum

Jules-Bertholet:
Averages the docking force between the two ports. This fixes the [docking port Kraken drive](https://wiki.kerbalspaceprogram.com/wiki/Docking_Port_Drive) and other unphysical behavior.


gotmachine:
Nice work, and thanks for the contribution.

However, could you please refactor this as a new, separate bugfix patch ?

While I agree this is a bug, and that fixing it by default would overall benefit to the vast majority of players, I would prefer to at least give the option of disabling it for the people wanting to mess around with the kraken drive thing, especially since it is wildly documented.

Jules-Bertholet:
@gotmachine  I've done as you requested.

PR #173, opened 2023-12-18
Increase TT-18A Launch Clamp EC generation

Clayell:
The stock TT-18A launch clamp has quite a low amount of electricity generation. (1 EC/s) In stock this isn't a huge issue, as there aren't many parts that use a lot of electricity when not 'deployed'. However, even in small mods like Cryo Parts, the tank cooling can use several EC/s with a medium-sized craft, and energy generation sources simply can't meet this for long. (energy storage can hold for the actual flight, but not for the period before flight)

Shouldn't change stock gameplay at all, but make slightly modded gameplay a bit easier. (some mods do have launch clamps with higher EC generation that work similar to stock)
Sorry if the formatting is incorrect, I'm new to MM.


gotmachine:
I understand the rationale here, but KSPCF has a rule of not touching anything in terms of balance, and more generally to avoid touching any stock configs.

PR #177, opened 2024-01-10
Set name for substitute part

Nazfib:
When the substitute part for the editor part tooltip window is created, the name of the part is not set. This means that the substitute part gets the default Unity name of "{prefabName}(Clone)". With this PR, the part name gets set correctly to the `AvailablePart`'s `name` field, as is done elsewhere where a new part is instantiated from the prefab (e.g. at https://github.com/KSPModdingLibs/KSPCommunityFixes/blob/a411542b0897770c59dce0408b2569d7200635f9/KSPCommunityFixes/BugFixes/UpgradeBugs.cs#L377-L380).

I noticed this when investigating an issue in RP-1: it shows an "Unlock requirements" entry in the tooltip, which (among other things) contains the part name. 

PartnameErrorExample

### DIFF
--- a/KSPCommunityFixes/BugFixes/PartTooltipUpgradesApplyToSubstituteParts.cs
+++ b/KSPCommunityFixes/BugFixes/PartTooltipUpgradesApplyToSubstituteParts.cs
@@ -100,6 +100,7 @@ namespace KSPCommunityFixes
                 }
                 Part p = GameObject.Instantiate(ap.partPrefab);
                 p.gameObject.transform.SetParent(_substitutePartRoot.transform, true);
+                p.name = ap.name;
                 p.partInfo = ap;
                 try
                 {


gotmachine:
This isn't a bug in KSP. Nobody is supposed to rely on the GameObject name to identify a part, and even less in an user facing string.

I'm reluctant to do such mostly unnecessary changes just to fix some (likely easily fixable) issue in one mod, as this might just as well break something in another.

Judging from the screenshot and the not very user friendly tech required shown with the tech internal name as well, I'd say the feature is likely a bit work in progress ?

Nazfib:
Fair enough, I wasn't aware that stock did the same thing. I'll fix it on RP-1's end.

gotmachine:
On a side note, it should perfectly fine to change the `GetInfo()` implementation to use :
- `part.PartInfo.title` instead of `part.name`
- `ResearchAndDevelopment.GetTechnologyTitle(part.partInfo.TechRequired)` instead of `part.partInfo.TechRequired`

PR #198, opened 2024-02-11
Fix #195: clone the srfAttachNode when reversing a surface attachment during re-root

### LINKED ISSUE #195: ReRootPreserveSurfaceAttach does not reconnect the child attachnode to the parent part
Ends up causing https://github.com/JonnyOThan/TweakScale/issues/56

Normally the attachedPart would be set in ReverseSrfNodeDirection

However, disabling the patch leaves the attachnode position in the wrong place - so the patch is good, just missing a step.

I'm not sure if this has any other possible negative effects, but it does persist across a save/load of the craft file.

JonnyOThan:
A different take on ReRootPreserveSurfaceAttach.  Slightly closer to the apparent intended behavior from stock.

### DIFF
--- a/GameData/KSPCommunityFixes/Settings.cfg
+++ b/GameData/KSPCommunityFixes/Settings.cfg
@@ -174,7 +174,10 @@ KSP_COMMUNITY_FIXES
 
   // Disable the stock behavior of altering surface attachment nodes on re-rooting, which is 
   // unnecessary and doesn't work correctly, leading to permanently borked attachement nodes.
-  ReRootPreserveSurfaceAttach = true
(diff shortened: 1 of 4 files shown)


JonnyOThan:
hrm there seems to be some extra changes in here that shouldn't be :/ let me see if I can clean that up

gotmachine:
Well, as I mentioned in the related issue, my opinion is that this is way too hacky and is very likely to cause weird side effect for any piece of code expecting a surface attached part to be attached by the surface attach node. 

Doing a quick GH search, I find way too many pieces of code doing checks of the `p.srfAttachNode.attachedPart == something` style for this patch to have no side effect. This even happen in some stock code.

Issue #206, opened 2024-02-25
Better Undo/Redo can merge two steps into one


numberlesstim:
Steps to reproduce:

1. Place 2 tanks attached to each other.
2. Move the second (not root tank) using the "Move" tool.
3. Undo step 2.
4. Observe the tank disappearing instead of getting moved to the original place.

Using Redo after this will add the tank back in at the position it was _after_ moving it, so it appears that two steps have been merged.

This "merging" of steps also happens if you change anything in the paw. For example changing fuellevel, open/close a bay, extend panels, change part variant etc.

I did test this with no other mods installed and verified this does not happen without KSPCF.

numberlesstim:
Further, if in the above steps you instead move the root tank, it will both remove the second tank and move the root tank back to the original place. Maybe that is a better example.

gotmachine:
So...

It's not really a "merging" of steps, but more an oversight of the consequences of inverting the undo/redo logic. 

Basically, what the patch does is invert the core undo/redo logic from capturing state *after* events to capturing it *before* events. 

However, I only moved at which point the state is captured for attach/detach events, not for offset/rotate events, which means you end up with twice the same state captured (in your reproduction steps, you will notice that you have to undo twice at step 3 for step 4 to happen), and the state before offsetting/rotating not being captured at all.

Now that I'm looking at it, there are actually more events that result in the same issue (same repro steps, just change step 2 with the following actions) :
- Switching between part variants
- Removing parts from symmetry
- Adding/removing part actions to action groups, and resetting part actions on the selected part
- Using the reset button of the stage manager

This will need quite a bit of work to fix them all. Basically, we need to move the call to `EditorLogic.SetBackup()` before the actions take place (instead of after). 

- [ ] Offset/Rotate gizmo
The backup is created from callbacks called from the gizmo components `OnMoveEnd` / `OnRotateEnd` methods, triggered when the gizmo is released. The gizmos do have a `On*Start` method, but unfortunately they aren't hooked to a callback, and those are generic components that can potentially be used elsewhere. We could still patch that method, and compare the instance to the `EditorLogic.gizmo*` fields. Another solution would be to call `SetBackup()` at the beginning of the editor FSM events `st_offset_tweak.OnEnter` and `st_rotate_tweak.OnEnter`, but this means entering the offset/rotate will always trigger a backup (instead of only if the part is actually moved), and that only one backup will be created (instead of every time the gizmo is mouse-released).

[...]

PR #210, opened 2024-03-02
Fixes for issue #206 (BetterUndoRedo)

gotmachine:
We need to move calls to `EditorLogic.SetBackup()` before the actions take place (instead of after), see issue #206

- [x] Offset/Rotate gizmo
The backup is created from callbacks called from the gizmo components `OnMoveEnd` / `OnRotateEnd` methods, triggered when the gizmo is released. The gizmos do have a `On*Start` method, but unfortunately they aren't hooked to a callback, and those are generic components that can potentially be used elsewhere. We can still patch those methods, and compare the gizmo instance to the `EditorLogic.gizmo*` fields.
- [x] Switching between part variants
~Requires moving the `SetBackup()` call at the beginning of the `ModulePartVariants.onVariantChanged` method~
Doesn't work because the field is already set when `ModulePartVariants.onVariantChanged`, so the changes will be saved to the backup anyway. Alternatively, we patch `UIPartActionVariantSelector.SelectVariant()`
- [x] Removing parts from symmetry
Requires moving the `SetBackup()` call at the beginning of the `Part.RemoveFromSymmetry` method
- [x] Adding/removing part actions to action groups, and resetting part actions on the selected part
Requires moving the `SetBackup()` call at the beginning of the `EditorActionGroups` `ResetPart`, `AddActionToGroup` and `RemoveActionFromGroup` methods
- [x] Using the reset button of the stage manager
~Requires moving the `SetBackup()` call at the beginning of the `StageManager.Reset` method~
Actually after further testing, it seems the backup is already created before the action, so nothing to fix here

### DIFF
--- a/KSPCommunityFixes/QoL/BetterEditorUndoRedo.cs
+++ b/KSPCommunityFixes/QoL/BetterEditorUndoRedo.cs
@@ -1,4 +1,5 @@
-﻿// see https://github.com/KSPModdingLibs/KSPCommunityFixes/issues/172
+﻿// see https://github.com/KSPModdingLibs/KSPCommunityFixes/issues/172 for initial implementation
+// and https://github.com/KSPModdingLibs/KSPCommunityFixes/pull/210 for following fixes
 
 // In stock, undo state is captured after part events are complete, and undoing will restore that state captured before that.
 // This make the user experience quite poor as undoing will loose all PAW tweaks made in between attach/detach actions.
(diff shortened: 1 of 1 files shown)


gotmachine:
Reverted to draft, as I forgot that we need to keep the firing of the `GameEvents.onEditorShipModified` from `SetBackup()` at the original point (ie, after the action took place). So more work is needed on those patches...

gotmachine:
@JonnyOThan feel free to merge and push to a release

PR #215, opened 2024-03-13
More memory leaks

JonnyOThan:


KSPCommunityFixes/Performance/MemoryLeaks.cs
```diff
@@ -580,6 +694,155 @@ static bool ModuleScienceExperiment_OnAwake_Prefix()
...
+                __instance._fi.occludersBodyCount = 0;
+                __instance._fi.occludersConvectionCount = 0;
+                __instance._fi.occludersSunCount = 0;
+            }
+
+            // not sure about
+            // objectUnderVessel
+
+            FlightGlobals.ResetObjectPartUpwardsCache();
+            FlightGlobals.ResetObjectPartPointerUpwardsCache();
```
gotmachine, line 798: Not ideal to reset that cache here since that will often cause it to be immediately rebuilt in every case a debris or another vessel gets out of loading range. Ideally we could be a bit smarter by waiting for next frame and only removing destroyed parts from those caches, but probably not worth the trouble, and I agree this is a potentially quite bad leak.

gotmachine:
LGTM

PR #230, opened 2024-06-03
FlightIntegrator and friends micro-optimization

gotmachine:
This is ongoing work to micro-optimize the various heavy duty stuff in FlightIntegrator, VesselPrecalculate and possible other flight scene low level components. The focus is on improving performance scaling in large part count situations, mainly by micro-optimizing the per-part work done in these component. 

The optimizations are meant to keep behavior as close to stock as possible, but they do introduce minor changes, and given how critical the patched methods are, any bug or behavior change is likely to have dire cascading consequences, so this will need some testing before being pushed.

As of now, there are a few other areas I want to look at before considering pushing this to a (pre)release, but I won't have much time to commit in the next weeks, not sure when I will resume working on this.

Here is a preview release : [KSPCommunityFixes_FlightPerf.zip](https://github.com/user-attachments/files/15537932/KSPCommunityFixes_FlightPerf.zip)

### DIFF
--- a/GameData/KSPCommunityFixes/Settings.cfg
+++ b/GameData/KSPCommunityFixes/Settings.cfg
@@ -420,6 +420,10 @@ KSP_COMMUNITY_FIXES
   // state synchronization and caching solar panels scaled space raycasts results.
   OptimizedModuleRaycasts = true
 
+  // General micro-optimization of FlightIntegrator and VesselPrecalculate, significantely increase
+  // framerate in large part count situations.
+  FlightPerf = true
+
   // ##########################
   // Modding
   // ##########################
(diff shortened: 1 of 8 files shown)


R-T-B:
> I think there's some kind of issue related to the camera anchor point. This started happening after I installed this build in tpksp: https://clips.twitch.tv/ModernBusyLemurTinyFace-fvxZ_jtUEAGkjCZc
> 
> Possibly involving docking or loading a save.

I may have seen this as well.   On the vessel in question, do you also have where you can't get back to the spacecenter via the top menu, but only vis esc and such?

R-T-B:
Ah, well those are my symptoms when this bug strikes, at any rate.

PR #232, opened 2024-06-18
Fix an IndexOutOfRangeException in DragCube.Load()

Nazfib:
The name field in a `DRAG_CUBE` config node is supposed to be optional. However, when the name is not included in the list of values, an `IndexOutOfRangeException` is thrown.

Even when not loaded from a config file, when the name field on a `DragCube` object is the empty string (for example, when it is default-constructed in code) it is not included in the string returned by `DragCube.SaveToString()`; this causes a problem when the `FlightIntegrator.Setup()` method uses this string to clone drag cubes.

The first instructions in `DragCube.Load(string[] data)` are:
 - Check if it has the correct length (24 or 25), logging an error and returning if this is not the case
 - Check if the length is 12; if it is not, take the first field as the name and skip it.

That last check should have been against a length of 24 instead; this patch replaces the first occurrence of `ldc.i4.s 12` in the code by `ldc.i4.s 24`.

I've checked KSP version 1.0.5 (the oldest version available via Steam), and the bug was already present in that version; the code seems to be unchanged between then and version 1.12.5.

### DIFF
--- a/GameData/KSPCommunityFixes/Settings.cfg
+++ b/GameData/KSPCommunityFixes/Settings.cfg
@@ -222,6 +222,9 @@ KSP_COMMUNITY_FIXES
   // cargo bays, wrong vessel size being reported, etc...
   PartBoundsIgnoreDisabledTransforms = true
 
+  // Fix loading of drag cubes without a name failing with an IndexOutOfRangeException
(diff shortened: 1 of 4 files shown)


JonnyOThan:
Thanks for the patch!

PR #235, opened 2024-06-21
PAWStockGroups - Generators

yalov:
Add Generator PAW Group, 
a separate group per every generator

     

Without groups, there is no information about which generator and action the fields are related to.

### DIFF
--- a/KSPCommunityFixes/QoL/PAWStockGroups.cs
+++ b/KSPCommunityFixes/QoL/PAWStockGroups.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using HarmonyLib;
 using KSP.Localization;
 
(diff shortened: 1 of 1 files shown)


yalov:
`ModuleResourceConverter` that have coretemp and termal efficiency is not the `ModuleGenerator`.
`ModuleResourceConverter` does not have the same problem:

gotmachine:
I would recommend adding a check to only put generators in groups if there are multiple generators on the part.

JonnyOThan:
Yeah agreed; and sorry I guess that should be filed as a new feature request rather than a comment on a PR - I still need to go over this in detail but it looks good.

yalov:
> I would recommend adding a check to only put generators in groups if there are multiple generators on the part.

@gotmachine , why? it groups action and fields together, and adds resources abbr

gotmachine:
> why? it groups action and fields together, and adds resources abbr

From the feedback I've got, people tend to dislike having to go through the extra step of having to expand groups, especially when the PAW doesn't have much stuff. In general, there is a lot of negative feedback on the whole "extra groups" patch. At least I would set those generator groups as always expanded by default, set true in the last arg, instead of the conditional on module count :
```cs
BasePAWGroup pawGroup = new BasePAWGroup(name, generatorGroupTitle + abbrs, __instance.part.Modules.Count > 3);
```

PR #250, opened 2024-09-06
Improve SceneLoadSpeedBoost patch

siimav:
Adds another change under the `SceneLoadSpeedBoost` umbrella. Forces all scene changes to happen synchronously which on some scenes can significantly lower load times. In my own testing with RP-1 I saw 4-4.5 seconds shaved off from SC -> VAB and VAB -> flight transitions. Others reported even 6-7s improvements.
The main speedup comes from asset cleanup running once instead of 3 times. However even if the async code path was forced to run cleanup only once, it was still close to 1s slower than the sync path.

### DIFF
--- a/GameData/KSPCommunityFixes/Settings.cfg
+++ b/GameData/KSPCommunityFixes/Settings.cfg
@@ -302,6 +302,7 @@ KSP_COMMUNITY_FIXES
   // ##########################
 
   // Reduce scene loading time by caching the current save in memory instead of loading it from disk
+  // Additionally forces all scene transitions to happen synchronously.
   SceneLoadSpeedBoost = true
 
   // Prevent the buoyancy integrator from eating CPU cycles when the vessel isn't submerged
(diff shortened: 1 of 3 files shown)


ValiusV:
Hi! That sounds amazing. Would it be possible to have the compiled fix ? Because the official release is unlikely to happen soon..

siimav:
I do have the same change available separately from KSPCF. Just drop the dll to the root of gamedata.
[NoLoadingBuffer.zip](https://github.com/user-attachments/files/17091478/NoLoadingBuffer.zip)

ValiusV:
Thanks ! It works really well and I haven’t encountered any problems yet. 

I conducted a few performance tests with my heavily modded modpack. I repeated them several times in the same situations, and here are the results:
-loading VAB: 12-13s without fix, 7-7.5s with fix
-loading KSC: 16-18s without fix, same with fix (no improvement in this scene)
-loading flight scene with a 100 parts ship: 36-38s without fix, 30-31s with fix

GameData/KSPCommunityFixes/Settings.cfg
```diff
@@ -302,6 +302,7 @@ KSP_COMMUNITY_FIXES
   // ##########################
 
   // Reduce scene loading time by caching the current save in memory instead of loading it from disk
+  // Additionally forces all scene transitions to happen synchronously.
```
JonnyOThan, line 305: This should be a separate cfg option - being able to toggle these independently is vital for identifying when one of them causes an issue.

Issue #251, opened 2024-09-22
Can we fix the problem of time warping through planets?


JonnyOThan:
When on a collision or aerobraking trajectory, it’s pretty easy to warp through a point where rails warp should have been disabled.  Can we do anything about this?  It shouldn’t be difficult to calculate a point on your orbit where you shouldn’t warp past.

gotmachine:
> shouldn’t be difficult

Famous last words :P 

To describe the full problem, what is needed is a way to detect if a point moving along an arbitrary 3D ellipse is intersecting a sphere moving along another arbitrary ellipse between a time interval, where those two objects have an arbitrary and non linear velocity over time.

But lets simplify the problem a bit. Assume that the vessel and bodies are moving in a straight line : the problem is now a relatively simple ray / capsule intersection check, which can be done by decomposing the check into two ray/sphere intersections and a ray/cylinder intersection.

But you have to consider that at the timesteps of the highest timewarp rates, a vessel or a body can move by a significant portion of a full orbit revolution, and in many cases, by more than a full revolution. If you just take the position of the vessel and of the body at the current fixed frame and at the next fixed frame to generate your ray and capsule respectively, that will generate garbage results. 

To have the ray and capsule be an acceptable representation of the elliptical trajectories, you need to decompose the time step into smaller ones where both orbits curvature between the start and end times is small enough, said otherwise, where the angle between the normalized velocity vectors at times t and t+n is small enough. There are probably smart ways to find those points, but a naive bisection is gonna take a lot of iterations. 

So, to summarize :
- For each vessel
- Discretize the next fixed time step into smaller steps approximating the orbit ellipse section into lines of acceptable error.
- For each body that it can potentially collide with
- Discretize its trajectory too
- For every vessel and body time pair, get their position and perform a ray/capsule intersection check.

Obviously, this would be quite computationally expensive, and definitely not compatible with decent frame rates if done naively using the KSP orbit stuff. There are maybe a few smart ways to early out of the full check and to make it faster, but I wouldn't qualify any of this as "not difficult".

JonnyOThan:
Ah, yeah this isn’t a complex math problem at all. The game already knows where the SOI translations are. Even within a single SOI you can warp through the body or atmosphere.  That’s a simple calculation to find the time at which you cross a certain altitude.

The hard part, I think, is making sure timewarp doesn’t go past that point.

gotmachine:
[...]

PR #252, opened 2024-09-25
Experimental fix for #251 (timewarp through bodies)

gotmachine:
This is preleminary work for issue #251, but I'm actually not sure what we're trying to fix.
Please discuss it on the issue, not here.


gotmachine:
A fix for this issue made by @JonnyOThan was released in 1.36

PR #257, opened 2024-10-07
FlightIntegrator and friends micro-optimization Round 2

JonnyOThan:
Many experimental perf improvements, basically a big collection of micro-optimizations for various pieces of code running every frame on every part. Won't do much in low part count situations, and some patches are highly situational (atmospheric flight, docking ports, struts, particles repositionning on floating origin shifts...)

Patches affecting the FlightIntegrator and VesselPrecalculate are broably incompatible with mods trying to override/extend stuff through MFI. There are a bunch of perf issues with "end of the call stack" methods being inefficient because they do extra work that is already done at the call site or can be made faster by using state from the call site, so patching those methods doesn't cut it, but this mean many stock methods aren't called anymore, so if those methods are replaced/prefixed/postfixed with MFI, things will break. Notable examples are FI.UpdateAerodynamics() and FI.Integrate(Part). This could probably be avoided at the cost of some of the perf gains, but I wanted to see how far things could be pushed perf wise. On a side note, there is a bug lurking in the VesselPrecalculate.CalculatePhysicsStats reimplementation, not sure what it is exactely but a side effect is camera target being wrongly offset. Likely something with the CoM calcs.

### DIFF
--- a/GameData/KSPCommunityFixes/KSPCommunityFixes.version
+++ b/GameData/KSPCommunityFixes/KSPCommunityFixes.version
@@ -2,7 +2,7 @@
   "NAME": "KSPCommunityFixes",
   "URL": "https://raw.githubusercontent.com/KSPModdingLibs/KSPCommunityFixes/master/GameData/KSPCommunityFixes/KSPCommunityFixes.version",
   "DOWNLOAD": "https://github.com/KSPModdingLibs/KSPCommunityFixes/releases",
-  "VERSION": {"MAJOR": 1, "MINOR": 35, "PATCH": 0, "BUILD": 0},
+  "VERSION": {"MAJOR": 1, "MINOR": 35, "PATCH": 1, "BUILD": 0},
   "KSP_VERSION": {"MAJOR": 1, "MINOR": 12, "PATCH": 5},
   "KSP_VERSION_MIN": {"MAJOR": 1, "MINOR": 8, "PATCH": 0},
   "KSP_VERSION_MAX": {"MAJOR": 1, "MINOR": 12, "PATCH": 5}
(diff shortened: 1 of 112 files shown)


JonnyOThan:
This is the other set of changes from https://github.com/KSPModdingLibs/KSPCommunityFixes/pull/230

gotmachine:
So, I feel this is ready for a (pre)release. 

Compatibility with MFI and relevant plugins using it (FAR, KerbalWeatherProject, AdvancedAtmosphereTools...) has been fixed, although not tested extensively, but I don't foresee issues here.

The various perf fixes have been reorganized into various "thematic" patches : 
- **ModuleDockingNodeFindOtherNodesFaster** : Faster lookup of other docking nodes.
- **CollisionEnhancerFastUpdate** : Optimization of the `CollisionEnhancer` component (responsible for part to terrain collision detection).
- **PartSystemsFastUpdate** : Optimization of various flight scene auxiliary subsystems : temperature gauges, highlighter, strut position tracking...
- **MinorPerfTweaks** : Various small performance patches (volume normalizer, eva module checks)
- **FlightIntegratorPerf** : General micro-optimization of `FlightIntegrator` and `VesselPrecalculate`, components in charge of most of heavy lifting for newtonian physics as well as atmospheric and thermal physics.
- **FloatingOriginPerf** : General micro-optimization of floating origin shifts. Main benefit is in large particle count situations (ie, launches with many engines) but this helps a bit in other cases as well.

There are two other (not really related) changes in this PR : 
- Small internal refactor of the patching infrastructure for less verbose patch declaration.
- Introduced a new "override" patch type, basically an automatic transpiler allowing to replace a method body with another. This has a little less overhead than a prefix doing the same thing, and allow for other patches (including non-KSPCF ones) to prefix the patched method as usual. This was surprisingly difficult to put together and to get right, and while I have done quite a bit of validation, it wouldn't surprise me if further issues arise with this. For now, this new patch type is only used by the patches in this PR, but if everything works out, updating all the existing "prefix return false" patches would be nice.

PR #259, opened 2024-10-09
Fix #251: Fix stock code that fails to limit timewarp when approaching a SOI transition

### LINKED ISSUE #251: Can we fix the problem of time warping through planets?

JonnyOThan:


JonnyOThan:
This works pretty well, but I want to test with alarms as well

JonnyOThan:
seems good.

gotmachine:
It shall be noted that this fix only works if "ORBIT_WARP_DOWN_AT_SOI" is true in `settings.cfg`. There is no UI for that setting, so probably not a big deal.

Other than that, there is still the issue that the stock "stop warp on trajectory crashes into body" implementation is reliant on the configured per-body altitude limits being adequately set for the max time warp rate (which can be vastly higher than stock). Ideally, it would nice to put together a revamp of the whole logic using sensible altitude defaults and relying on checking against the intersection UT instead of against the altitude at discrete points.

JonnyOThan:
Right...I think I'll play around with WarpEverywhere to construct a repro case and make sure your patch (or something similar) works there.

gotmachine:
To be clear, I think your patch is good enough. It effectively fixes the main practical issue the we ended up identifying, going though a body lying in the next SOI. The rest are likely more theoretical issues than practical ones.

PR #269, opened 2024-10-16
Loading optimizations

gotmachine:
Note that this PR depends on / include #272 

Various loading performance focused changes, part of the general loading re-implementation :
- Added performance metrics logging upon reaching the main menu
- Added (almost) entirely custom MU model parser, roughly 3 times the throughput of the stock parser (~300 MB/s on my machine). Will only really benefit to people having fast NVME drives with good random read performance.
- KSPCF now maintains dictionaries of loaded models and texture assets by their url/name, and patch the stock `GameDatabase.GetModel*` / `GameDatabase.GetTexture*` method to use them instead of doing a linear search.
  This was especially bad with models, as the method would compare the requested string to the `GameObject.name` property for every model in the database.
  Unfortunately, we have no way to know if additional assets are loaded by custom means, so we have to fallback to a linear search when the requested texture isn't known and we can't assume that a texture that isn't found at some point won't be found in the future, so in cases where a not loaded / absent texture is queried, there is no performance improvement.
  Overall, this benefit to many scenarios, during initial loading for model parsing and part compilation, and more marginally for scenes switches [when various initialization paths are re-acquiring a texture reference](https://github.com/search?q=GameDatabase.Instance.GetTexture+OR+GameDatabase.Instance.GetModel+language%3AC%23&type=code).

Other changes, also loading related :
- As a part of the `MinorPerfTweaks` patch, patched the `FlightGlobals.fetch` property to not fallback to a `FindObjectOfType()` call when the `FlightGlobals._fetch` field is null, which is always the case during loading. In a stock + BDB test case, this alone was about 10% of the total loading time, 7+ seconds. The call doesn't seem necessary as `FlightGlobals._fetch` is set/unset from `FlightGlobals.Awake()` / `FlightGlobals.OnDestroy()`. I guess there are some very edge cases where a (just about to be destroyed) instance could be acquired by a call to `FindObjectOfType()`, but in any case I would qualify such behavior as a bug. On a side note, the same issue is present with `PhysicsGlobals`, but I haven't been able to detect any benefits in patching that one.
- New patch `PartParsingPerf`
  - Slightly faster part icon generation. Part icon GameObject generation is done by cloning the part prefab and basically removing everything on it. The operation represent roughly 35% of the part compilation time, a good two thirds of that being in instantiating the copy of prefab, and running all the associated code (part / modules KSPField scaffolding and Awake() code), only to immediately destroy everything. Unfortunately, this can't really be avoided due to the need to run `PartModule.OnIconCreate()` and in some case, some model-manipulation code in Awake().
  - Faster `Part` fields parsing, by creating a dictionary of IL-emitted parser delegates instead of the very generic reflection based stock approach.

Overall, these changes can provide a quite significant boost to loading time, mainly to part compilation and model loading. On a hot boot (config/model/texture loading not throughput limited by I/O), total time from exe launch to main menu for a stock + BDB install is now around 35 seconds on my 5800X3D.

As usual, this probably need to be tested a bit more before being released.


KSPCommunityFixes/Performance/FastLoader.cs
```diff
@@ -557,10 +731,15 @@ static IEnumerator FastAssetLoader(List<ConfigFileType> configFileTypes)
 
             // start model loading
             gdb.progressFraction = 0.75f;
+            KSPCFFastLoaderReport.wTextureLoading.Stop();
+            KSPCFFastLoaderReport.wModelLoading.Start();
             gdb.progressTitle = "Loading model assets...";
             yield return null;
 
             // call non-stock model loaders
+            modelsByUrl = new Dictionary<string, GameObject>(allModelFiles.Count);
```
Cgettys, line 740: Why not
```
modelsByUrl = new Dictionary<string, GameObject>(allModelFiles.Count, StringComparer.OrdinalIgnoreCase);
modelsByDirectoryUrl = new Dictionary<string, GameObject>(allModelFiles.Count, StringComparer.OrdinalIgnoreCase);
```
And the same for texturesByUrl?
It should have the same semantics as your existing equality logic, but probably will be faster based on your comment elsewhere about frequently falling through.
It might be slightly slower in the case where it's an exact match,  but not drastically - and it should eliminate fallback to the O(n) case for case insensitive matches. I don't know if you still need the fallback (I saw in some cases that you might have to check a List because you could just directly modify it in stock KSP code for one of these optimizations), but thought I'd mention it.
https://learn.microsoft.com/en-us/dotnet/api/system.stringcomparer.ordinalignorecase?view=net-8.0
https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.-ctor?view=net-8.0#system-collections-generic-dictionary-2-ctor(system-int32-system-collections-generic-iequalitycomparer((-0)))

KSPCommunityFixes/Performance/PartParsingPerf.cs
```diff
@@ -0,0 +1,666 @@
...
+                iconRenderers.Clear();
+                materialBuffer.Clear();
+            }
+
+            Vector3 size = partBounds.size;
+            float x = Math.Abs(size.x);
+            float y = Math.Abs(size.y);
+            float z = Math.Abs(size.z);
+            iconScale = x > y ? x : y;
+            iconScale = iconScale > z ? iconScale : z;
```
Cgettys, line 207: curiosity: does Math.max get optimized differently by the compiler here than this expression?

KSPCommunityFixes/Performance/FastLoader.cs
gotmachine: Model search is case sensitive, but yeah that could make sense for textures.

TBH, occurrences of the methods being called with mismatching casing were pretty rare in my tests (less than 1%), and for the few cases where it happens, the requested casing will be added as duplicate entry, so the overhead will only occur on the first call for a specific casing. The case insensitive comparer will still induce quite a bit of overhead, so I'm not so sure that would be a net gain overall.

Falling back to a linear search on the original stock lists is necessary in any case (:P), as (a few) models/textures are also loaded from asset bundles and because there nothing preventing external code from modifying the lists directly. That still leave a hole if someone decide to *remove* models/textures from them, but that feel unlikely enough.

On a side note, upon further review, we should probably also patch `GetTextureInfoIn()` and `GetModel()`, as those are commonly used too, including during part compilation.

KSPCommunityFixes/Performance/PartParsingPerf.cs
gotmachine: Surprisingly, even a function as small as `Math.Max()` is not inlined by default by the rather terrible JIT compiler used by Unity/KSP, so it actually does matter, at least in hot paths.

This method is far from a hot path, but I guess I've taken the habit of manually inlining stuff like that even when not necessary, which is poor practice as it make the code a bit less idiomatic and harder to read. I should add my own versions of such methods with the aggressive inlining attribute, that would be better.

KSPCommunityFixes/Performance/FastLoader.cs
Cgettys: Fair - I assumed it was significantly more frequent based on this comment on the issue (which I happened across by chance when being curious about what the latest version you just published held :)).

"The dictionary approach is however not as beneficial here, because those methods allow to get a texture by its case-insensitive url, so we have to fallback to iterating on the list with a case insensitive string comparison if the requested url isn't correctly cased. The net result was still beneficial in my tests, but not as much as I hoped **due to that slow path being taken relatively often.**"

InvariantCultureIgnoreCase and CurrentCultureIgnoreCase are fairly expensive; I'd be surprised (but would believe you) if OrdinalIgnoreCase was particularly expensive though. Then again, my knowledge may be more relevant for newer dotnet versions? (ex: https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-8/).

But fair, you're trading an initial miss + a bit of memory in an uncommon case, to save a bit of arithmetic / bit-twiddling + possibly other overheads on every case - I'm not a KSP modding expert, your intuition is likely better than mine in this area and you're in a position to measure it for sure :).

Pity you can't get rid of that silly list - I was gonna suggest "well if you can modify, why not binary search" - but that won't work if code makes additions without being aware of the need to insert at the right position.

gotmachine: Yeah, unfortunately, the mono-derived BCL and JIT compiler used in KSP/Unity is missing all the juicy improvements from more or less the last 10 years (and in many cases, perform just plain worse than the good old .Net framework).

This being said, the case-insensitive dictionary still make sense and is a trivial change, so I will try and check what the profiling results have to say.

gotmachine:
So, for reference, here are some profiling results regarding the GameDatabase methods :
This was tested on a stock + DLC + BDB + Tantares/TantaresLV + SXT + KWRocketry + a few NearFuture mods, for a total of 2676 parts and 142 IVAs.

The texture getting methods were called 13 326 times during loading : 

- Stock : 
image
- KSPCF with the default comparer. The slow path was taken 26 times (0.22%) : 
image
- KSPCF with the OrdinalIgnoreCase comparer. The slow path was never taken : 
image

So all in all, I'd say that the difference is pretty negligible, but using the default comparer is either faster or just as fast. I haven't verified but I would guess the majority of the slow path cases happened in `GetTextureInfo()` (and this is likely given the usage of it). The figures might start to invert a bit when the slow path is taken more often, but this feel unlikely.

The real gains are for models anyway : 
- Stock : 
image
- KSPCF : 
image

KSPCommunityFixes/Performance/FastLoader.cs
Cgettys: Ah, right... I haven't touched Unity in about half a decade. My frame of reference for "old dotnet performance" is .net framework... I forgot that Unity has its own fork of Mono.
And I'm assuming Unity's version of  Mono is just different enough / just tightly coupled enough that we can't e.g. just drop in a modern version of the dotnet runtime either... Though it might be fun to try sometime :smile:

KSPCommunityFixes/Performance/PartParsingPerf.cs
Cgettys: Makes sense. The reason it stood out  to me was because you only did it for Math.Max
e.g. `Math.Abs(size.y)` could have been `size.y >= 0.0 ? size.y : -size.y`
And so on :)

Given the JIT you have to work with, I can see the argument for doing it. Though as you said, if the compiler will happily inline your own Math utility class so long as you do```[MethodImpl(MethodImplOptions.AggressiveInlining)]```, that's the best of both worlds.

Pity we can't (or maybe we could, I'd have to think about it) peephole optimize some of this in the IL before the Unity Mono JIT gets its hands on it.

Cgettys:
> So, for reference, here are some profiling results regarding the GameDatabase methods : This was tested on a stock + DLC + BDB + Tantares/TantaresLV + SXT + KWRocketry + a few NearFuture mods, for a total of 2676 parts and 142 IVAs.
> 
> The texture getting methods were called 13 326 times during loading :
> 
>     * Stock :
>       image
> 
>     * KSPCF with the default comparer. The slow path was taken 26 times (0.22%) :
>       image
> 
>     * KSPCF with the OrdinalIgnoreCase comparer. The slow path was never taken :
>       image
> 
> 
> So all in all, I'd say that the difference is pretty negligible, but using the default comparer is either faster or just as fast. I haven't verified but I would guess the majority of the slow path cases happened in `GetTextureInfo()` (and this is likely given the usage of it). The figures might start to invert a bit when the slow path is taken more often, but this feel unlikely.
> 
> The real gains are for models anyway :
> 
>     * Stock :
>       image
> 
>     * KSPCF :
>       image

I'd agree, thanks for sharing the numbers / satisfying my curiousity. And it's an interesting technique I hadn't though of for handling case insensitive lookup where _usually_ the casing will match but it won't always.
This code is single threaded, right? Otherwise some synchronization is probably required (or ConcurrentDictionary, but in my experience that's pretty expensive).

KSPCommunityFixes/Performance/PartParsingPerf.cs
Cgettys: > Pity we can't (or maybe we could? I know you're making changes to IL already, just not sure what the restrictions are and why they are / whether you're doing it in place, putting in thunks, or what) peephole optimize some of this in the IL before the Unity Mono JIT gets its hands on it.

Ah... that's what "transpiler" patches are (I was thinking one language to another, not  IL to IL).
Maybe I'll try that sometime, if you haven't already. But agreed that MethodImplOptions or manual inlining is better for code you're patching in, rather than having to transpile the patched code too.

PR #281, opened 2024-10-31
Optimise VectorLine camera projection for improved Map View performance

Halbann:
Orbit rendering takes up a significant majority of the frame time with a large number of orbit lines visible in the Map View, which is mainly caused by camera projection done by VectorLine.

This patch improves performance by caching the camera's projection matrix multiplied with its view matrix once per frame, and replacing calls to different camera projection functions in VectorLine with our cached versions.

Will write a full description with some benchmarking results when the patch is more complete.


KSPCommunityFixes/KSPCommunityFixes.csproj
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
```
Halbann, line 1: I don't know why this line is marked as changed.

KSPCommunityFixes/Performance/OptimisedVectorLines.cs
```diff
@@ -0,0 +1,261 @@
...
+            DontDestroyOnLoad(this);
+        }
+
+        protected void Update()
+        {
+            // I think this is faster than it would be to compare Time.frameCount to a cached frame count in every call to VectorLineCameraProjection.
+
+            VectorLineCameraProjection.cacheDirty = true;
+        }
+    }
```
gotmachine, line 260: Better try to avoid using specialized KSPAddons whenever possible.
The main KSPCommunityFixes `KSPAddon` already has a static fixed frame counter `long` incremented from its `FixedUpdate()`, you can add an `Update()` there incrementing a `long` frame counter and use that to compare to the last cached frame. The performance difference between comparing a long and a bool is really insignificant here.

KSPCommunityFixes/Performance/OptimisedVectorLines.cs
```diff
@@ -0,0 +1,261 @@
...
+            m_Viewport = camera.pixelRect;
+
+            // Instead we use these.
+            m_worldToCameraInv = camera.worldToCameraMatrix.inverse;
+            m_projectionInv = camera.projectionMatrix.inverse;
+
+            cacheDirty = false;
+        }
+
+        private static Vector3 WorldToClip(Matrix4x4 m, Vector3 point)
```
gotmachine, line 229: Pass the matrix by ref here. (large) struct copying is really expensive.

KSPCommunityFixes/Performance/OptimisedVectorLines.cs
```diff
@@ -0,0 +1,261 @@
...
+    public static class VectorLineCameraProjection
+    {
+        // Based on CameraProjectionCache from UnityCsReference.
+        // https://github.com/Unity-Technologies/UnityCsReference/blob/2019.4/Editor/Mono/Camera/CameraProjectionCache.cs
+
+        public static bool patchEnabled = true;
+        public static bool cacheDirty = true;
+
+        static Matrix4x4 m_WorldToClip;
+        static Matrix4x4 m_WorldToClipInverse;
```
gotmachine, line 113: It should be worthwhile to use a double backed Matrix4x4 here.

The Mono version KSP uses actually convert all floats to double internally, every time a basic operator is used. So every time a non-trivial amount of float operations are performed (which is typically the case with matrix <> vector math), it is hugely faster to convert stuff to double upfront. We have a 3x4 double backed matrix implemented in Numerics.cs for such purposes, but here I believe you need the full 4x4 matrix, so I would recommend taking the time to implement it (could be a good addition to Numerics.cs as well).

KSPCommunityFixes/Performance/OptimisedVectorLines.cs
```diff
@@ -0,0 +1,261 @@
...
+            };
+
+            // Perspective multiplication moves it to clip space.
+            clipSpace.x *= clipSpace.w;
+            clipSpace.y *= clipSpace.w;
+            clipSpace.z *= clipSpace.w;
+
+            Vector4 viewSpace = m_projectionInv * clipSpace;
+            viewSpace.w = 1f;
+            Vector4 worldSpace = m_worldToCameraInv * viewSpace;
```
gotmachine, line 196: Related to the above comment about using a double backed 4x4 matrix, I would recommend using a vector4d (passing it by ref) here to avoid intermediary float to double conversions, and mutating a single vector4d it instead of instantiating new ones. 

Instantiating new structs for intermediary results is actually pretty expensive.

KSPCommunityFixes/KSPCommunityFixes.csproj
gotmachine: It also happens to me pretty frequently, IDEs fighting over file encoding I guess ? doesn't really matter...

gotmachine:
I would recommend validating your results by comparing them to the original method results.
You can gate that validation code behind a file level `#if def`, and use ULP based equality comparisons (`Numerics.AlmostEqual()`, with maybe an ULP of 5 in this this case), see an example here : https://github.com/KSPModdingLibs/KSPCommunityFixes/blob/93b57a17289b4ce7139873d88c3a209af2781718/KSPCommunityFixes/Performance/FlightIntegratorPerf.cs#L684-L729

KSPCommunityFixes/KSPCommunityFixes.cs
```diff
@@ -26,6 +26,9 @@ public class KSPCommunityFixes : MonoBehaviour
 
         public static long FixedUpdateCount { get; private set; }
 
+        // Frame counter that doesn't use an external call like Time.frameCount.
+        public static long frameCount;
```
Halbann, line 30: Is there any chance that using a field instead of a property here saves performance given how often this is called?

gotmachine: AFAIK, simple properties are always be inlined as a read over their backing fields, so that shouldn't matter.

PR #285, opened 2024-11-30
Allow targeting the currently orbited parent body, or any body the parent body is a descendant of

jamespglaze:
# Problem
In stock KSP, you are unable to target your direct parent body, or any body that the parent body is a descendant of.  This primarily crops up when attempting to plan gravity assist chains, particularly when entering resonant orbits that will return to the initial body.  Some examples of where this can be annoying is in planning of the following assist chains starting from Kerbin:
1. Kerbin(Mun-Mun-Mun)-Kerbin(Mun)-Kerbin(Mun)-Eve
2. Kerbin-Eve-Kerbin-Kerbin-Jool
3. Kerbin-Leverage-Kerbin-Jool

In each case, it can be difficult to set up the first Kerbin flyby without first ejecting from Kerbin blind, since you cannot target Kerbin to see your future close approach.
This limitation is not a result of an inability of the targeting logic to target bodies in the parent hierarchy, but rather dedicated logic for dropping the target if it is.

# Changes
* Use a prefix to skip calling `OrbitTargeter.DropInvalidTargets` entirely.  The only function of this method is to drop the target if it is the current parent or any body the parent body is a descendant of.
* Use a transpiler to modify `FlightGlobals.UpdateInformation`.  One of the primary functions of this method is to check the validity of the current target and drop it if invalid.  The first two branches of the set of if statements call `FlightGlobals.SetVesselTarget` to drop the target if it is the current parent or any body the parent body is a descendant of.  These first two if statements are entirely replaced with Nops.

# Testing
I have taken the following steps to test the changes made
* Run profiling using the unity profiler to confirm there are no performance regressions due to this change.
* Rollback KSP version to 1.8.0 to confirm compatibility.
* Performed numerous ingame tests using the functionality for planning assist chains.
* Used the changes in the course of normal gameplay over the past few weeks.

I have observed no bugs or performance regressions with the latest version of the changes, and additionally have tested that they function alongside Kerbal Engineer Redux targeting behavior.  I have not tested with MechJeb2 targeting behavior.

### DIFF
--- a/GameData/KSPCommunityFixes/Settings.cfg
+++ b/GameData/KSPCommunityFixes/Settings.cfg
@@ -308,6 +308,9 @@ KSP_COMMUNITY_FIXES
   // Optional MM-patcheable toggle to always disable the MH features
   OptionalMakingHistoryDLCFeaturesAlwaysDisable = false
 
+  // Allow targeting the parent body of the current craft, or any body in the parent hierarchy
+  TargetParentBody = true
+
   // ##########################
   // Performance tweaks
   // ##########################
--- a/KSPCommunityFixes/KSPCommunityFixes.csproj
+++ b/KSPCommunityFixes/KSPCommunityFixes.csproj
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="..\packages\Krafs.Publicizer.2.2.1\build\Krafs.Publicizer.props" Condition="Exists('..\packages\Krafs.Publicizer.2.2.1\build\Krafs.Publicizer.props')" />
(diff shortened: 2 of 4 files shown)


Clayell:
The amount of times I have been annoyed by being unable to target your parent body is incredible and certainly helped me transition to Principia. I can attest that this PR would be very helpful to add into KSPCF, perhaps even defaulted to on.

gotmachine:
I can definitely see how this is useful, however I'm a bit concerned about potential issues with mods relying on such a behavior not being possible in stock. I guess the stock behavior is mainly here to prevent targeting the sun / root body (which is much more likely to break stuff), but that still doesn't seem possible with your changes, at least when targeting stuff with the mouse. 

I think it would probably be safer to still have checks in place to drop the root body, in case it is set as a target through the public API by something else, but well...

As usual, we won't really know if this can cause issues until we push this publicly, so lets go for it I guess. I quickly checked MechJeb, it seems to have enough guards in place in the various displays and planners to handle this somewhat gracefully.

PR #289, opened 2024-12-29
Support cleaned dll files from de4dot or binary patchsets

R-T-B:
So, I've been testing what patches do/don't break with KSPCF and a de4dot cleaned dll (as the cleaned dll performs ever so slightly better than the obfuscated one).

I've pretty much made a patchset that can completely support all current patches with the cleaned dlls and no errors.  I use SHA256 hash checks against my own binary patches [here](https://github.com/R-T-B/KAC) as well as a size detection coupled with version check to do some heuristics to detect other homebuilt cleaned-dll builds.

I can keep it separate in my own repos of course, but wanted to make you aware in case you thought this worthy of mainlining.  Take a look and tell me what your thoughts are, no offense taken if you reject it.

### DIFF
--- a/KSPCommunityFixes/BugFixes/DoubleCurvePreserveTangents.cs
+++ b/KSPCommunityFixes/BugFixes/DoubleCurvePreserveTangents.cs
@@ -11,7 +11,38 @@ namespace KSPCommunityFixes.BugFixes
 
         protected override void ApplyPatches()
         {
-            AddPatch(PatchType.Prefix, typeof(DoubleCurve), nameof(DoubleCurve.RecomputeTangents));
+            if (!KSPCommunityFixes.cleanedDll)
+            {
+                AddPatch(PatchType.Transpiler, typeof(DoubleCurve), nameof(DoubleCurve.RecomputeTangents));
+            }
+            else
+            {
+                AddPatch(PatchType.Prefix, typeof(DoubleCurve), nameof(DoubleCurve.RecomputeTangents));
+            }
+        }
+
+        static IEnumerable<CodeInstruction> DoubleCurve_RecomputeTangents_Transpiler(IEnumerable<CodeInstruction> instructions)
+        {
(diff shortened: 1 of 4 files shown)


gotmachine:
I'm not a huge fan of having duplicate implementations, and even less having them applied selectively by checking magic numbers such as file size.
 
I'm also skeptical all the other transpilers that this PR doesn't touch are working as intended on deobfuscated codebases, most of them are likely to just fail silently to apply.

I agree the current situation is a bit unfortunate. Ideally, we should review all transpilers and rework them to be more robust and work seamlessly on both obfuscated and deobfuscated versions of the code, but I don't have the motivation for such an undertaking...

R-T-B:
"I'm not a huge fan of having duplicate implementations, and even less having them applied selectively by checking magic numbers such as file size.

I'm also skeptical all the other transpilers that this PR doesn't touch are working as intended on deobfuscated codebases, most of them are likely to just fail silently to apply.

I agree the current situation is a bit unfortunate. Ideally, we should review all transpilers and rework them to be more robust and work seamlessly on both obfuscated and deobfuscated versions of the code, but I don't have the motivation for such an undertaking..."

We talked about this earlier and I agree.  My fork remains unadvertised in the forums for that reason, until I (or someone) personally review the transpilers.

As for the "magic numbers" I trued using sha256 hash first but that fails because the hash is different for every cleanup done by de4dot.

gotmachine:
There is also the fact that is only really a minor inconvenience for plugin developers, there is zero reason for end users to run a deobfuscated game...

R-T-B:
Yeah, I weighed that as well and its partially why I stopped work on this.  I don't really have plans for advertising or updating it in the near or even distant future, I honestly forgot I had submitted a PR.  The repo will remain up in case it proves useful to someone for furthering this work for whatever reason, but I agree, neither you or should dedicate any more resources to this.  Thanks for reviewing anyways.

PR #292, opened 2025-02-05
enable commnet throttling by default

JonnyOThan:


gotmachine:
See #299 as a more balanced solution, that will be enabled by default.

Issue #306, opened 2025-03-30
Fuel Delivery Overlay in editor not showing fuel flows


ngx-ree:
### What did you do before the bug occurred?

Using 'Fuel Delivery Overlay' button in VAB editor either on tanks or on engines.

### What happened?

Fuel Delivery Overlay function in VAB/SPH editor does not show fuel flows, only nodes on engines and tanks. This bug is present since pre-release 1.37.0. Previous versions seem ok.

### Upload your ksp.log file

[KSP.zip](https://github.com/user-attachments/files/19526382/KSP.zip)

### A screenshot of the problem

![Image](https://github.com/user-attachments/assets/f6251cc6-220d-4d6d-af8e-4d902716fb66)
![Image](https://github.com/user-attachments/assets/6b3a3055-dd04-4850-9192-1278a327f157)

gotmachine:
Thanks for the report.
The `OptimisedVectorLines` patch is responsible for this bug, haven't investigated further.

Will look into it ASAP, in the meantime you can disable that specific patch with the following MM patch :
```
@KSP_COMMUNITY_FIXES:FINAL
{
  @OptimisedVectorLines = false
}
```

ngx-ree:
Thank you for reply and workaround. The bug is not a big issue, overlay is still usable as nodes appear only on parts between which the flow is possible. It's just less obvious without lines showing, which is inconvenient especially with more complex vessels.

gotmachine:
The results of the `ScreenToWorldPoint()` method KSPCF implementation are definitely wrong : https://github.com/KSPModdingLibs/KSPCommunityFixes/blob/342a0692ecceaf5ba9238ea798148dc70200fc40/KSPCommunityFixes/Performance/OptimisedVectorLines.cs#L212-L230

I managed to get the overlay to appear by commenting this line, which is applying this part of the patch : https://github.com/KSPModdingLibs/KSPCommunityFixes/blob/342a0692ecceaf5ba9238ea798148dc70200fc40/KSPCommunityFixes/Performance/OptimisedVectorLines.cs#L40

But even without that there is still some weird offset between where the overlay appear and where it should be : 

![Image](https://github.com/user-attachments/assets/66b5a4a4-2aed-46b9-9272-b9ef73b2f22a)

 Forcing a cache update on every call doesn't solve it, so I'm guessing the base projection / matrix math we are using is wrong to some extent.

@Halban Can you look at this ?

[...]

PR #315, opened 2025-05-07
Fix projection mistake in Optimised Vector Lines

Halbann:
While we were optimising our WorldToScreen/Viewport functions, we lost the [intended behaviour](https://docs.unity3d.com/6000.1/Documentation/ScriptReference/Camera.ScreenToViewportPoint.html#:~:text=right%20is%20(1%2C1).-,The%20z%20position%20is%20in%20world%20units%20from%20the%20camera.,-using%20UnityEngine%3B) of returning the signed (negative is behind the camera) world distance from the camera plane in the z component, and introduced a discrepancy between our functions and Unity's functions where the world position is projected as though it were `nearClipPlane` units closer to the camera plane than it actually is. 

It's hard to see in-game given the size of orbits in scaled space relative to `nearClipPlane`, but I would have noticed this if I'd written those tests like @gotmachine suggested in the PR.

The issue with the z component leads to a noticeable bug where placing the camera close to a 2D orbit line, which is possible for the orbit lines of planets, causes the final line segment before the near clip plane to point somewhere off-screen in the wrong direction.

I don't see any issues related to the ScreenToWorld function so I've left that as is.

https://github.com/user-attachments/assets/f477d531-ea0c-43c1-8ca5-c3ff0155bd45

https://github.com/user-attachments/assets/50a37dfc-0ba3-4609-ad00-e7ad32774f6e

### DIFF
--- a/KSPCommunityFixes/Performance/OptimisedVectorLines.cs
+++ b/KSPCommunityFixes/Performance/OptimisedVectorLines.cs
@@ -91,12 +91,12 @@ namespace KSPCommunityFixes.Performance
             return ReplaceCall(instructions, Camera_ScreenToWorldPoint, VectorLineOptimisation_ScreenToWorldPoint, count);
         }
 
-        private static IEnumerable<CodeInstruction> ReplaceScreenToViewportPoint(IEnumerable<CodeInstruction> instructions, int count)
+        private static IEnumerable<CodeInstruction> ReplaceViewportToWorldPoint(IEnumerable<CodeInstruction> instructions, int count)
         {
-            MethodInfo Camera_ScreenToViewportPoint = AccessTools.Method(typeof(Camera), nameof(Camera.ScreenToViewportPoint), new Type[] { typeof(Vector3) });
-            MethodInfo VectorLineOptimisation_ScreenToViewportPoint = AccessTools.Method(typeof(VectorLineCameraProjection), nameof(VectorLineCameraProjection.ScreenToViewportPoint));
+            MethodInfo Camera_ViewportToWorldPoint = AccessTools.Method(typeof(Camera), nameof(Camera.ViewportToWorldPoint), new Type[] { typeof(Vector3) });
+            MethodInfo VectorLineOptimisation_ViewportToWorldPoint = AccessTools.Method(typeof(VectorLineCameraProjection), nameof(VectorLineCameraProjection.ViewportToWorldPoint));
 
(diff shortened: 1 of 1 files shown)


gotmachine:
Does this fully solve issue #306 ?
`ScreenToWorldPoint` is seemingly problematic as well there.

Halbann:
No, afraid not. I'll fix that one asap. My handle on github is @Halbann rather than Halban.

Halbann:
The fuel overlay works now and the results look identical in Unity. I'm going to ping Charlie and see if he wants to take a crack at simplifying the new version or if he only needs to make a small tweak to fix the old version.

Halbann:
I thought I'd write those checks that you suggested last year to make sure we don't have to come back to this again. It turned out that while they look correct, they're not nearly as accurate as you wanted (ULP <= 5), or at least they don't match Unity's functions  to that precision.

This is from the tracking station with no vessels, just planet orbit lines. The average number of correct (ULP < maxUlps) and incorrect results is taken over 100 frames at each level of maxUlps.

|maxUlps|averageCorrect               |averageIncorrect|Average Percent Correct|
|-------|-----------------------------|----------------|-----------------------|
|5      |830.7901                     |165.21          |83.41                  |
|7      |924.7201                     |71.28           |92.84                  |
|10     |951.93                       |44.06999        |95.58                  |
|15     |967.8499                     |28.15           |97.17                  |
|20     |973.5699                     |22.43001        |97.75                  |
|30     |979.05                       |16.95           |98.3                   |
|40     |982.1602                     |13.84           |98.61                  |
|50     |984.1                        |11.9            |98.81                  |
|100    |987.3201                     |8.68            |99.13                  |
|200    |990.62                       |5.379998        |99.46                  |
|300    |993.3201                     |2.680001        |99.73                  |
|400    |994.54                       |1.46            |99.85                  |
|500    |995.18                       |0.8199999       |99.92                  |
|1000   |995.9                        |0.09999999      |99.99                  |
|2000   |996                          |0               |100                    |
|5000   |996                          |0               |100                    |
|10000  |996                          |0               |100                    |

image

I don't really know what to do with that information other than to say that while it's correct now, it's not as accurate (or rather as similar to Unity) as you expected and I'm not sure why. ScreenToWorldPoint is generally more accurate than WorldToScreen, and WorldToScreen and WorldToViewport are about the same. I haven't noticed any wildly incorrect results.

gotmachine:
1000 ULPs of error isn't that big considering the how many operations are compounded there.
Results are looking okay to me.

PR #329, opened 2025-08-02
New ksp bug fix patch : DebugConsoleDontStealInput

gotmachine:
Fix the Alt+F12 console input field stealing input when a console entry is added


Clayell:
Darn, does this mean [DebugStifle](https://github.com/KSP-CKAN/NetKAN/pull/10665) is redundant now? Just released that too...

I was thinking yesterday that this could be a KSPCF thing lol.

PR #336, opened 2025-10-04
Small simplification of FlightIntergratorPerf.SetDragCubeDrag

CensoredUsername:
This probably shaves a few cycles, and if nothing else makes the code easier to understand.

I found this while working on a [mod that attempts to fix some odd drag cube behaviour](https://github.com/CensoredUsername/BodyDragFix). Not sure if that is a qualifier for KSPCommunityFixes, it does change the stock behaviour quite a bit. Basically what it does is, instead of lerping on `dot` it lerps on `dot*dot`.

Unfortunately this does also result in a compatibility issue between KSPCommunityFixes and my mod right now. Would anyone have some advice on how to deal with this?


CensoredUsername:
>but as you are the only one whose mod will be broken by changing it I figure I will leave it up to you.

No worries, in the end I learned enough about C# bytecode patching so I wrote (and tested) my mod so it should be compatible with either version. Feel free to merge it!

PR #338, opened 2025-10-23
perf: Optimize `MonoUtilities.RefreshContextWindows`

Phantomical:
This calls `UnityEngine.Object.GetObjectsOfType` which is terribly slow. It is not used anywhere in stock which is probably why it hasn't been patched yet. However, several mods do make use of it.

On my KSP instance this saves > 1s on every scene switch across a few different mods.

The improvement here is to maintain our own list of active `UIPartActionWindow` instances and use that list instead of calling `GetObjectsOfType`. It proactively removes all dead instances when one is deleted. Any change that results in `UIPartActionWindow.OnDestroy` not being called will result in a memory leak anyways so I suspect this is a non-issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ##### Unreleased
 **New/improved patches**
 - **MinorPerfTweaks** [now includes a patch to make PQS loading faster](https://github.com/KSPModdingLibs/KSPCommunityFixes/pull/343) on scene switches, saving ~1 second when switching from KSC to a vessel.  Thanks @Phantomical!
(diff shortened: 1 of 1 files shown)


gotmachine:
Pretty sure there can only be single PAW per part.

Yeah, it would make sense to use the same implementation for `MonoUtilities.RefreshPartContextWindow()` as well.

Those `MonoUtilities` methods are pretty ancient and predate the addition of the `Part.PartActionWindow` property, I think it was introduced sometimes between KSP 1.5 and 1.8.

PR #341, opened 2025-11-03
Fix lower limit for wheel moment of inertia

MajorNr01:
When the game calculates the moment of inertia for wheels, it sets a hard-coded lower limit of `0.01`. This is much larger than the calculated moment of inertia of most landing gear wheels (e.g. the LY-10 Small Landing Gear has a moment of inertia of `0.000648`).

This causes significant forces on small aircraft (~10t) during touchdown as the wheels spin up. As a result, the craft pitches down if the wheels are not spinning and pitches up if the wheels are still spinning. This is correct in principle but very exaggerated in small aircraft due to the inappropriately large moment of inertia.

This patch changes the lower limit to `0.00001` which is lower than the correct moment of inertia of the stock landing gears.

The chosen lower limit is kind of arbitrary (I just added three zeroes) and I am completely open to lowering it even further or removing it entirely.

### DIFF
--- a/GameData/KSPCommunityFixes/Settings.cfg
+++ b/GameData/KSPCommunityFixes/Settings.cfg
@@ -240,6 +240,11 @@ KSP_COMMUNITY_FIXES
   // Fix the Alt+F12 console input field stealing input when a console entry is added
   DebugConsoleDontStealInput = true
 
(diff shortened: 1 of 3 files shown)


Phantomical:
Thanks for the PR!

PR #350, opened 2026-01-04
Avoid duplicate call to PSystemSetup.SetActive in FlightDriver.setStartupNewVessel

Phantomical:
When you launch out of the VAB, FlightDriver.setStartupNewVessel ends up starting all PQS instances twice. The initial start of PQS is very expensive, and all the work done on the first call is basically thrown away.

This patches `FlightDriver.setStartupNewVessel` to avoid calling `PSystemSetup.SetActive` a second time. This is just a matter of patching the 3 different calls to shims that track whether they have been called.

How much this actually ends up saving depends on the planet and mods installed. For stock I expect this to be about 2s, for the Sol install I'm testing (with a whole bunch of other patches to various mods) it is about 2.5s.

### Before
image

### After
image

### DIFF
--- a/GameData/KSPCommunityFixes/Settings.cfg
+++ b/GameData/KSPCommunityFixes/Settings.cfg
@@ -340,6 +340,9 @@ KSP_COMMUNITY_FIXES
   // and on other occasions, wasting a ton of CPU processing time.
   PQSCoroutineLeak = true
 
+  /// Prevent KSP from restarting PQS for the current planet multiple times when launching a new ship.
+  PQSOnlyStartOnce = true
+
   // Remove unused ProgressTracking update handlers. Provides a very noticeable performance uplift in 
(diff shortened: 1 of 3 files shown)


siimav:
Does it help in practice (i.e outside the profiler) too? With Plain Old RSS the results with this patch disabled vs enabled are within measurement error.

Phantomical:
Some quick testing on a (mostly) stock shows that this is definitely happening. I don't think I have anything installed on this new testing save that would affect the results here. Looking at the code this should happen any time you launch from the VAB or SPH and the launchsite is a PQSCity or a PQSCity2. @siimav what did your test setup look like?

image

Comparison with patch enabled

image

The launch does feel noticeably faster to me - which makes sense since it seems to nearly be a 4s improvement. I'll configure up a fully stock + KSPCF install and give it another test.

PR #363, opened 2026-04-12
Add support for custom debugging builds.

R-T-B:
This simply adds support for custom dll builds where possible, mostly by using prefixes or fixed transpilers.  It uses a basic size heureustic that is pretty reliable to detect custom dlls of Assembly-CSharp.

This is mostly useful for debugging, and is expected to be best effort, not a truly supported thing, but I expect it would be welcomed by some.

### DIFF
--- a/KSPCommunityFixes/BugFixes/DoubleCurvePreserveTangents.cs
+++ b/KSPCommunityFixes/BugFixes/DoubleCurvePreserveTangents.cs
@@ -11,7 +11,7 @@ namespace KSPCommunityFixes.BugFixes
 
         protected override void ApplyPatches()
         {
-            if (!KSPCommunityFixes.cleanedDll)
+            if (!KSPCommunityFixes.IsCleanedDll)
             {
                 AddPatch(PatchType.Transpiler, typeof(DoubleCurve), nameof(DoubleCurve.RecomputeTangents));
             }
--- a/KSPCommunityFixes/BugFixes/DoubleCurvePreserveTangents.cs
+++ b/KSPCommunityFixes/BugFixes/DoubleCurvePreserveTangents.cs
@@ -1,7 +1,5 @@
 ﻿using System;
-using System.Collections.Generic;
-using HarmonyLib;
-using System.Reflection.Emit;
+using System.Runtime.CompilerServices;
 
 namespace KSPCommunityFixes.BugFixes
 {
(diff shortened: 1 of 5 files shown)


R-T-B:
Should be ready now, there were some slight transpiler logic errors on MSVC built dlls but I fixed them.

KSPCommunityFixes/KSPCommunityFixes.cs
```diff
@@ -52,6 +52,35 @@ public static Version KspVersion
...
+                {
+                    Byte[] data = File.ReadAllBytes(dllPath);
+                    if ((data.Length < 10485760) && (KSPCommunityFixes.KspVersion >= new Version(1, 12, 0)))
+                    {
+                        return true; //certainly a home-cleaned dll, no official 1.12.x build of Assembly-CSharp is less than 10MBs.
+                    }
+                }
+                return false;
+            }
+        }
```
Phantomical: This reads the entirety of Assembly-CSharp every time it is queried.
1. It should be cached.
2. It should use FileInfo to get the length of the file instead of reading it.
3. It should use `typeof(GameDatabase).Assembly.Location` to get the location of `Assembly-CSharp` instead of relying on platform-dependent paths

R-T-B:
converting to draft pending more testing.

KSPCommunityFixes/BugFixes/ModuleIndexingMismatch.cs
```diff
@@ -222,6 +220,7 @@ private static void LoadProtoPartSnapshotModule(ProtoPartModuleSnapshot protoMod
             protoModule.moduleRef = module;
         }
 
+
         static void LoadModules(ProtoPartSnapshot protoPart, Part part)
```
Phantomical, line 224: The assembly sequences for this method are virtually identical other than the `dup; pop` pair. It should be possible to merge the two branches here. I'm also not entirely clear why the opcodes are being overwritten with `nop` instead of just being removed, but that looks like it was there already.

KSPCommunityFixes/KSPCommunityFixes.cs
Phantomical: Also, nit, but this should be called something like `IsCleanedDll`

Phantomical (commented):
I will probably have more comments but I need to do my taxes so I'm posting these for now.

KSPCommunityFixes/KSPCommunityFixes.cs
R-T-B: Fixed in latest commit.

KSPCommunityFixes/BugFixes/ModuleIndexingMismatch.cs
R-T-B: I'm not actually comfortable removing the branching structure.  I tried earlier and it failed miserably.  Note the "cleaned dll" branch features two fingerprint styles as well, not one, complicating things (Mono vs MSVC outputs).

If anyone else wants to do that they are welcome, but this is the best I can pull off for the moment, so marking resolved as this whole PR is only for "best effort" support for cleaned dlls.

R-T-B:
So, I think it's ready now.  But welcome any further comments.

The worst part is probably how it uses file size to ID stock dlls, but, well, it certainly won't ever do the wrong codepath on a stock user, and the support for modded dlls is explicitly understood to be best effort, so...

R-T-B:
That should be it for tweaks, I have made a test build here and will be asking custom deobfuscated dll / debugging build users to play with it and see what happens.

https://github.com/R-T-B/KSPCommunityFixes/releases/tag/1.40.1_RTBFORKr2

KSPCommunityFixes/BugFixes/DoubleCurvePreserveTangents.cs
```diff
@@ -37,5 +44,44 @@ static IEnumerable<CodeInstruction> DoubleCurve_RecomputeTangents_Transpiler(IEn
 
             return code;
         }
+
+        static bool DoubleCurve_RecomputeTangents_Prefix(DoubleCurve __instance)
```
Phantomical: Since we're unconditionally overriding the method this should be `PatchType.Override`, and the method renamed to suit. It's a bit more efficient and you can then remove the bool return.

KSPCommunityFixes/BugFixes/RefundingOnRecovery.cs
```diff
@@ -56,7 +58,14 @@ class RefundingOnRecovery : BasePatch
...
         protected override void ApplyPatches()
         {
-            AddPatch(PatchType.Transpiler, typeof(Funding), "onVesselRecoveryProcessing");
+            if (!KSPCommunityFixes.IsCleanedDll)
+            {
+                AddPatch(PatchType.Transpiler, typeof(Funding), "onVesselRecoveryProcessing");
+            }
+            else
+            {
+                AddPatch(PatchType.Prefix, typeof(Funding), "onVesselRecoveryProcessing");
```
Phantomical: Ditto here, should be an override patch.

Phantomical (commented):
Two more small comments. I'm also not quite happy with the ModuleIndexingMismatch one, so I'm going to take a shot at rewriting some of it to work appropriately.

KSPCommunityFixes/BugFixes/DoubleCurvePreserveTangents.cs
R-T-B: done.

Phantomical:
Alright, check your PRs for simplifications to the two patches that still use the conditional.

R-T-B:
> Alright, check your PRs for simplifications to the two patches that still use the conditional.

done and merged, thank you.

I'll test later today with my release build from MSVC just to be sure it works properly.

R-T-B:
The ModuleIndexingMismatch change fails with MSVC builds, at least.  I suspect it probably works fine with mono builds, since that is what you tested against.

I'll try to supply you with an example dll soon of that style for comparison.  The changes are usually pretty minor.

Everything else seems good.

R-T-B:
FYI I am now on the fence about the utility of this...

The whole point of a debugging build is predictability for debugging right?  Whats the utility in using lightly tested fragile patches in such an environment, honestly, when you can just disable patches?

It just kind of hit me like that.  Feel free to head check my thought process here, it was fun regardless lol.

Phantomical:
This is part of why I have been pushing for patches that are the same in both environments (well that and maintenance burden).

IMO the patches are useful _if they work_. Unfortunately, since you are the one with the weird environment the work to make sure that they actually work in that environment falls on you.

R-T-B:
The irony is I'm actually not the one with the weird environment, it's siimav I just got told how to replicate the process heh.  I do like to support him because he does great work but I need to make sure our support work is ALSO both useful and reliable.

I guess we can merge this if it does not bother you but it just seems so fragile without a real dedicated tester ( I won't honestly be doing that beyond these spot tests).

PR #364, opened 2026-04-15
Fix radiation flux being inconsistent during timewarp

siimav:
The problem with original code is that ptd.expFlux and ptd.unexpFlux get mutated during every thermal integration pass. This will cause the magnitude of the flux to increase with every timewarp step. 1x speed and analytic mode (>100x) behaved correctly before and are thus unchanged.

### DIFF
--- a/KSPCommunityFixes/BugFixes/FIUpdateRadiation.cs
+++ b/KSPCommunityFixes/BugFixes/FIUpdateRadiation.cs
@@ -8,7 +8,7 @@ namespace KSPCommunityFixes
 
         protected override void ApplyPatches()
         {
-            AddPatch(PatchType.Prefix, typeof(FlightIntegrator), "UpdateRadiation");
+            AddPatch(PatchType.Override, typeof(FlightIntegrator), "UpdateRadiation");
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


KSPCommunityFixes/BugFixes/FIUpdateRadiation.cs
```diff
@@ -0,0 +1,59 @@
...
+
+namespace KSPCommunityFixes
+{
+    public class FIUpdateRadiation : BasePatch
+    {
+        protected override Version VersionMin => new Version(1, 12, 0);
+
+        protected override void ApplyPatches()
+        {
+            AddPatch(PatchType.Prefix, typeof(FlightIntegrator), "UpdateRadiation");
```
Phantomical: Since we're unconditionally overriding the method this should be `PatchType.Override` (and you'll need to rename the method below to `FlightIntegrator_UpdateRadiation_Override`)

Phantomical (commented):
One thing to change here but otherwise LGTM.

KSPCommunityFixes/BugFixes/FIUpdateRadiation.cs
siimav: Like this?

Phantomical: Yep! Just like that :)

PR #369, opened 2026-04-30
Optimize and rewrite texture loading

Phantomical:
KSPCF's texture loader is pretty fast so normally I wouldn't really bother to optimize it, but I would like to do some additional work here and refactoring texture loading to go through asset bundles will help later on.

Here's what's changed:
* Texture loading has been refactored pretty significantly. Each texture load now gets its own coroutine.
* We now build an asset bundle on a background thread and use that to load all dds and cached png textures. This points the streaming data to the actual texture files on disk so unity takes care of reading the textures and uploading them to the gpu entirely off the main thread.
* For other textures DDS loading is now more async:
  * Create a texture
  * Wait a frame for unity to finish its first upload
  * Use AsyncReadManager to read directly into the array returned by GetRawTextureData()
  * Apply (and make it unreadable if needed)
* PNG/JPG textures are still loaded through UnityWebRequest, normal swizzling is now offloaded to a background thread. Cached png textures have their assets loaded via the asset bundle path above.
* TGA/MBM/etc. remains mostly the same, though normal swizzling is offloaded.

This still needs a proper performance evaluation but my unscientific tests show performance to be similar to what was there before.


Phantomical:
The performance comparisons above are inaccurate. Marking this as draft until I can come up with better ones (and those better ones actually show an improvement)

PR #374, opened 2026-05-10
perf: Cache some expensive operations in stock drill modules

Phantomical:
All 3 of these spend a significant amount of time performing operations that could be shared. This introduces caches for those operations where applicable.

The two things cached here are:
* ResourceMap.GetAbundance calls - these are expensive but also do not change for the same (vessel, harvester type, resource name) key. We can do them once per frame and then keep reusing the result.
* CheckForImpact - this does an expensive raycast and gets called twice per FixedUpdate. We cache this so it only gets called once.

On net this seems to make things about 50% faster for `ModuleResourceHarvester`. I expect the other two to have a fairly minor improvement, though I haven't profiled them.

Before:
image

After:
image

### DIFF
--- a/GameData/KSPCommunityFixes/Settings.cfg
+++ b/GameData/KSPCommunityFixes/Settings.cfg
@@ -512,6 +512,10 @@ KSP_COMMUNITY_FIXES
   // Load the expansion bundles in the background while performing part compilation.
   ExpansionBundlePreload = true
 
+  // Cache some expensive computations in stock resource converter and drill modules.
+  // This has a fairly significant impact on ships with lots of drills.
+  ResourceConverterPerf = true
+
   // ##########################
   // Modding
   // ##########################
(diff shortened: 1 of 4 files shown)


GameData/KSPCommunityFixes/Settings.cfg
```diff
@@ -512,6 +512,10 @@ KSP_COMMUNITY_FIXES
   // Load the expansion bundles in the background while performing part compilation.
   ExpansionBundlePreload = true
 
+  // Cache some expensive computations in stock resource converter and drill modules.
+  // This has a fairly significant impact on ships with lots of drills.
```
Clayell, line 516: Maybe best to rewrite this as something like:
`// Cache some expensive computations in stock resource converter and drill modules. Most noticeable on ships with lots of drills.` (in keeping with the template above with CraftBrowserOptimisations)
Right now it looks like the patch has a `negative` impact on ships with lots of drills

PR #384, opened 2026-05-26
Allow targetting a body with a future encounter in the map view

### LINKED ISSUE #381: Cannot left-click or double-click on a planet in map view if you have a maneuver node with a future intercept
In this save file:

There is one vessel with a kerbin->Duna transfer node that has an encounter with duna.

Image

Left-clicking or double-clicking on Duna in map mode does nothing at all, which makes it impossible to set Duna as the target.  It works normally for all other bodies.

Interestingly, left-clicking or double-clicking on Duna's orbit line itself works.

Deleting the maneuver node also resolves the problem.

[persistent.zip](https://github.com/user-attachments/files/27971667/persistent.zip)

Phantomical:
OrbitTarget.TargetCastModes ignores any orbit whose reference body is equal to refPatch.referenceBody. This was meant to prevent targeting the body whose SOI the vessel is currently in. When a maneuver node exists this makes it so you are unable to target the reference body for the last patch, making its icon unclickable.

This changes it so it compares against the active vessel's current main body instead of refpatch.

Fixes #381


JonnyOThan:
Is there any reason to exclude the vessel's current body?  There are definitely tons of cases where you want to focus the camera on the body you're currently orbiting.  Or maybe I'm misunderstanding the change.

SofieBrink:
Pretty sure this isn't referring to the camera, but the actual targeting mechanism. And i don't think it makes any sense to be able to target the orbiting body.

JonnyOThan:
This bug affects being able to double click on a body in map mode to focus on it. That’s important even when in Kerbin orbit.  Setting it as target is kinda weird, but there is another patch that allows setting ancestor bodies as target because sometimes it is useful. And if you do set it as target, seeing DN/AN relative to the equator can be useful - though I’m not sure if that would automatically happen without more work.

Phantomical:
This method actually is used for both targeting and focusing, AFAICT. There's an existing patch for allowing you to target the current vessel's parent body, and otherwise stock will delete the target. I think it's fine to just disable this part of the check unconditionally.

PR #416, opened 2026-07-25
Add audio clip dictionary for GameDatabase

tobiasnmf:
This pr adds dictionary lookup for audio clips, which speeds up loads especially with [RocketSoundEnhancement](https://github.com/KSPModStewards/RocketSoundEnhancement/tree/master/GameData/RocketSoundEnhancement), saving around 7-9 seconds for me with this mod.

The overridden methods are implemented similarly to the others in `GameDatabasePerf`, and like them there's an assumption that mods won't modify nor remove audio directly from the public `databaseAudio`. Audio added to `databaseAudio` directly and audio removed properly with `RemoveAudioClip` works fine. I searched for public code on GitHub which modify or remove from the `databaseAudio` field directly, and couldn't find anything.

### Modded profiles
<details>
<summary>Mods used when profiling</summary>
<pre>
KSP version: 1.12.5.3190

Installed modules:

\- ALCOR 1.0.1
\+ ASETAgency v2.0.2
\- ASETAvionics v3.0.1
\+ ASETProps v2.0.7
\- Astrogator v1.0.2
A Atmosphere (unmanaged)
\- AvalanchePlumes v1.0.5.1
\+ B9PartSwitch v2.21.0.4
\- BackgroundResourceProcessing 0.3.2
A BreakingGround-DLC 1.7.1 (unmanaged)
A BuildManager (unmanaged)
\- BurstPQS 0.1.24
A CelestialShadows (unmanaged)
\- Chatterer 0.9.99
\- ChattererExtended 0.6.2
A CityLights (unmanaged)
\- ClickThroughBlocker 1:2.1.10.23
\- CommunityCategoryKit v112.0.1
\+ CommunityResourcePack v112.0.1
\+ CommunityTechTree 1:3.4.5
\- ContractConfigurator v2.13.3.0
\- DE-IVAExtension v1.3.0
\- Deferred 1.3.5.0
\- DistantObject v2.2.1.7
\+ DistantObject-default v2.2.1.7
\- DockingPortAlignmentIndicator 6.13.1.0
\- EasyVesselSwitch 2.3
\- EditorExtensionsRedux 3.4.7
A EVEManager (unmanaged)
\+ FilterExtensions 3.2.9.1
\- FilterExtensionsDefaultConfig 3.2.9.1
\- Firefly 1.0.6
\+ FireflyAPI 1.0.0.0
\- Firespitter v7.17
\+ FirespitterCore v7.17
\+ FirespitterResourcesConfig v7.17
[...]

PR #419, opened 2026-07-28
Fix GetAudioClip fallback result bug

tobiasnmf:
I double checked my recent pr https://github.com/KSPModdingLibs/KSPCommunityFixes/pull/416 and noticed the result from the fallback check doesn't actually return the result. This pr fixes that bug.

### DIFF
--- a/KSPCommunityFixes/Performance/GameDatabasePerf.cs
+++ b/KSPCommunityFixes/Performance/GameDatabasePerf.cs
@@ -141,7 +141,7 @@ namespace KSPCommunityFixes.Performance
             {
                 // Fallback in case audio is added after indexing
                 List<AudioClip> audioClips = gdb.databaseAudio;
-                for (int i = audioClips.Count - 1; i >= 0; i--)
+                for (int i = audioClips.Count; i-- > 0;)
                 {
                     if (audioClips[i].name == url)
                     {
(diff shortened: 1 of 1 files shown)


KSPCommunityFixes/Performance/GameDatabasePerf.cs
```diff
@@ -151,6 +151,9 @@ static AudioClip GameDatabase_GetAudioClip_Override(GameDatabase gdb, string url
                     }
                 }
 
+                if (result != null)
+                    return result;
+
                 KSPCFFastLoader.audioByUrl.Remove(url);
                 return null;
```
Phantomical: Why don't you delete the whole section?
```cs
                if (result != null)
                    return result;

                KSPCFFastLoader.audioByUrl.Remove(url);
                return null;
```

That would make this match the other methods. I find the remove call to be quite weird here.

Phantomical (changes requested):

KSPCommunityFixes/Performance/GameDatabasePerf.cs
tobiasnmf: I tried something different at first, and when I changed the approach to be similar to the other methods I didn't realize that it didn't make any sense to keep that code. So I agree that the remove call is quite weird.

Phantomical (approved):

Issue #410, opened 2026-07-14
PR #390 is breaking calculations dependent on part size in the SPH.


BrettRyland:
### What did you do before the bug occurred?

PR https://github.com/KSPModdingLibs/KSPCommunityFixes/pull/390 (which was added in v1.41.0) is breaking BDArmory HP and armour calculations in the SPH by returning 0-sized parts when they're selected from the menu.
In general, it breaks any part size calculations in the editor performed from `OnStart`.

This is easily demonstrated using a test class
```csharp
public class PartSizeTest : PartModule
{
	public override void OnStart(StartState state)
	{
		var bounds = part.GetRendererBounds();
		Debug.Log($"DEBUG {part.partInfo.name} has {bounds.Length} bounds: {string.Join(", ", bounds)}, part size: {CalculatePartSize(bounds)}");
	}

	public Vector3 CalculatePartSize(Bounds[] bounds)
	{
		Bounds result = new Bounds(part.transform.position, Vector3.zero);
		for (int i = 0; i < bounds.Length; i++)
		{
			result.Encapsulate(bounds[i]);
		}
		return result.size;
	}
}
```
and patch (so that the class is added to all parts for the test)
```
@PART[*]
{
	%MODULE[PartSizeTest]
	{}
}
```
then starting with a MK1Fuselage tank and selecting a second one from the menu
<img width="378" height="411" alt="Image" src="https://github.com/user-attachments/assets/ca9c2ecd-51c7-4b68-bb6e-dc921a518066" />

With v1.40.1:
```
[LOG 13:35:23.511] DEBUG MK1Fuselage has 1 bounds: Center: (0.0, 10.0, 0.0), Extents: (0.6, 0.6, 1.0), part size: (1.3, 1.3, 1.9)
[LOG 13:35:24.723] DEBUG MK1Fuselage has 1 bounds: Center: (16.5, 1.2, 15.6), Extents: (0.6, 0.6, 1.0), part size: (1.3, 1.3, 1.9)
```
With v1.41.1 (or v1.41.0):
```
[LOG 13:33:34.369] DEBUG MK1Fuselage has 1 bounds: Center: (0.0, 10.0, 0.0), Extents: (0.6, 0.6, 1.0), part size: (1.3, 1.3, 1.9)
[LOG 13:33:36.939] DEBUG MK1Fuselage has 0 bounds: , part size: (0.0, 0.0, 0.0)
```

### What happened?

[...]

PR #423, opened 2026-08-08
Fix condition to ignore TransparentFX in pr390 to fix issue #410.

BrettRyland:
PR #390 introduced a condition to ignore TransparentFX when calculating part renderer bounds.

> This updates the patch to exclude said renderers. There are cases in the editor where everything is on the TransparentFX layer (detached parts) so we can't do this unconditionally, but it should work in all the cases that matter.

The condition used is insufficient and breaks mods that rely on those bounds for their calculations #410. E.g., BDArmory uses these as part of its hitpoint calculations when the part is created (i.e., while the part is detached).

This PR modifies the condition for when to ignore TransparentFX to take into account detached parts by checking if all parents of the part transform are on the TransparentFX layer.

### DIFF
--- a/KSPCommunityFixes/BugFixes/PartBoundsIgnoreDisabledTransforms.cs
+++ b/KSPCommunityFixes/BugFixes/PartBoundsIgnoreDisabledTransforms.cs
@@ -52,7 +52,7 @@ namespace KSPCommunityFixes
             //
             // This matches the same rule as done in https://github.com/KSPModdingLibs/KSPCommunityFixes/issues/150
             // which isn't really perfect, but is probably good enough for our purposes.
-            bool ignoreTransparentFX = !p.frozen;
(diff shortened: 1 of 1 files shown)


Phantomical:
So the problem with the original patch runs a bit deeper than what this approach will fix. The editor will occasionally put everything on the TransparentFX layer and anything that attempts to compute the bounds at that point (and possibly afterwards, not sure) will have issues.

I don't think just adding the check here will work. I do have an approach that I think will work well enough, but the branch for that is on my home computer and I won't be back until later this week. I'll make a PR then.

BrettRyland:
OK, good to hear that you have another approach that you think will be better.
The additional check that this PR adds was sufficient to fix the issue in my testing with BDArmory, where we need the part size to calculate HP and other details during `OnStart` or shortly afterwards, during which the part is typically fully on the TransparentFX layer in the editor.

PR #424, opened 2026-08-13
Fall back to an uncrewed control source on vessel separation

djungelorm:
`Vessel.FallBackReferenceTransform()` picks the control point for a vessel created by decoupling, undocking or a part being destroyed, but it asks `ShipConstruction.findFirstCrewablePart()` for it, which only accepts a part with crew capacity, crew actually aboard and control source. A probe core or an empty command pod fails that test, so the new vessel gets no control point at all and ends up oriented by its root part, breaking the navball, SAS and autopilots until the player sets one manually.

Override the method so that when `findFirstCrewablePart()` finds nothing, it falls back to the first part that is a control source, searched in the same root-first order. A crewed part still wins, and `findFirstCrewablePart()` is left alone for its other callers.

This was originally found via the kRPC mod, where programmatically decoupling a vessel would leave the newly created vessel in an invalid control state, and its controlling part would need to be set to the probe core on the new vessel.


Phantomical (approved):
It will be a bit before I'm back home and have time to merge new KSPCF PRs, but this LGTM.

PR #426, opened 2026-08-15
Stop stock toolbar buttons being duplicated on scene switch

djungelorm:
When switching game scenes quickly, stock toolbar icons can become duplicated. This patch fixes that.

`UIApp.AddToAppLauncher()` keeps the button it is given in a single field, `appLauncherButton`, which `UIApp.OnDestroy()` uses to give it back. Nothing stops an app going through it twice: `onGUIApplicationLauncherReady` fires more than once for a scene loaded while another is still being torn down, and since the coroutine waits `AppStartFrameDelay` frames before doing anything, both runs are in flight before either has a button. The second overwrites the field, so `OnDestroy` gives back only the second button and the first is left on the toolbar with nothing that can remove it.

The apps' own duplicate checks compound it. All of them (`ResourceDisplay`, `ActionGroupsApp`, `ContractsApp`, `MessageSystem`, `KnowledgeBase`, `MissionsApp` and `ApplicationLauncher`) test `Instance != null` rather than `Instance != this`, and `AddToAppLauncher` runs `OnAppInitialized`, so an app added twice finds the instance it set itself and destroys itself. Its singleton is left pointing at a destroyed object, which causes a flood of `NullReferenceException` in the log.

With this patch, an app now gets one trip to the toolbar at a time, and only while it is not already on it. A genuine second instance still destroys itself exactly as before, and an app whose button went with a torn-down launcher is still free to add itself again.

Tested in KSP 1.12.5 over repeated scene switches: `ResourceDisplay` buttons went from growing by two per entry into flight (past ten and still climbing) to peaking at two and settling at one, and the `KSP.UI` exceptions above went from ~15 per scene change to none.

Found via the kRPC mod, to drive rapid scene changes and so hits this on nearly every transition, but can accumulate in ordinary play too.


Phantomical:
Thanks for the fix!

PR #430, opened 2026-09-06
UIFloatEditNumericInput: don't write PAW fields on numeric mode toggle or for destroyed parts

sparr:
Disclosure: This PR was primarily authored by Claude Opus 5. I handed this task off when I encountered this bug while working on my own mod, asking the AI to reduce the bug to test cases and a fix. I am working around the bug in my mod, but suspect an upstream fix may be desirable. If AI-generated code is not welcome for merging here, please feel free to instead interpret this as just a bug report with attached proof-of-concept code for illustrative purposes.

### Problem

`UIPartActionNumericFloatEdit.Setup()` subscribes to a global game event and only unsubscribes from `OnDestroy()`:

```cs
GameEvents.onPartActionNumericSlider.Add(ToggleNumericSlider);
```

`onPartActionNumericSlider` fires once, for everyone, when "#" is clicked in any PAW. Every float edit item that has run `Setup()` and not yet run `OnDestroy()` receives it — not only the items of the window that was clicked. `ToggleNumericSlider(false)` then unconditionally calls `OnFieldInput(inputField.text)`, which goes through `UIPartActionFieldItem.SetFieldValue()` and so runs the module's `UI_Control.onFieldChanged`. Two problems follow.

**1. Turning "#" off writes fields nobody edited, at reduced precision.** `ToggleNumericSlider(true)` fills the box with `GetFieldValue().ToString($"F{floatControl.sigFigs}")`. Turning it back off parses that *already rounded* text, snaps it up to `incrementSlide` and writes it back. A "#" on/off cycle with no typing replaces every float edit field on screen with its own displayed rounding and runs all their change callbacks. A `sigFigs = 2` field holding 1.23456 comes out at 1.23.

**2. The write can land on a part that no longer exists.** The subscription outlives the part in ordinary situations: `Object.Destroy` defers `OnDestroy` to end of frame, and Unity never calls `OnDestroy` at all on a component whose `GameObject` was never activated. Turning "#" off then runs a destroyed module's change callback. Stock modules mostly shrug; module code that rebuilds meshes or walks the part's transforms does not:

```
[ERR] Exception handling event onPawNumericSlider in class UIPartActionNumericFloatEdit:
      System.NullReferenceException
    ROLib.ROLModelModule`1[U].UpdateModelScalesAndLayoutPositions (System.Boolean doNotRescaleX)
    ROLib.ModuleROTank.UpdateModelMeshes ()
    ROLib.ModuleROTank.ModelChangedHandler (System.Boolean pushNodes)
    ROLib.ModuleROTank.OnDiameterChanged (BaseField f, System.Object o)
    UIPartActionFieldItem.SetFieldValue (System.Object newValue)
    KSPCommunityFixes.UIPartActionNumericFloatEdit.OnFieldInput (System.String input)
    KSPCommunityFixes.UIPartActionNumericFloatEdit.ToggleNumericSlider (System.Boolean numeric)
    EventData`1[T].Fire (T data)
```

Any module whose `onFieldChanged` touches its part is a candidate.

### Reproducing

*Precision loss* — one part: get a float edit field to a value finer than its displayed precision, open its PAW, click "#" and click it again without typing. The value has changed to the rounded figure and `onFieldChanged` has run.

*Write to a dead part* — open a PAW on a part with a float edit field, get rid of the part (delete it, or leave the scene), then click "#" on any other PAW. The removed part's module gets its change callback.

### Fix

Two guards in `ToggleNumericSlider`:

- Return early, and drop the subscription, when the item, its part or its module has been destroyed.
- Only call `OnFieldInput` when `inputField.text` differs from the current value formatted the way `ToggleNumericSlider(true)` formats it — i.e. only when the box was actually edited. `RemoveInputfieldLock()` still runs on both paths.

The stock item this one is modelled on never writes the field at all — `UIPartActionFloatRange.ToggleNumericSlider` sets the three `SetActive` states, returns early if `!inputField.gameObject.activeInHierarchy`, and assigns `inputField.text`. It cannot fire `onFieldChanged`, and it already bails out for items that aren't on screen. So the write-back being conditioned here has no stock counterpart for anything to have depended on.

### Testing

KSP 1.12.5, modded install, driving the "#" toggle from an automated editor session. An instrumented build logging every item that receives the event shows it directly:

```
toggle    numeric=False field=currentLength part=<deadpart> selfDead=False partDead=True moduleDead=True activeInHierarchy=False
toggleOff              field=currentLength part=<deadpart> text='1.5000' cur='1.5000' differs=False
```

The item's own `GameObject` is alive, which is why it reaches the write; its part and module are destroyed; and the text matches the field at display precision, yet `SetFieldValue` still saw a change and ran `onFieldChanged` — the stored float was not exactly the displayed `1.5000`. That is the silent rounding above, caught causing the crash.

That same short run sent 38 of 69 event deliveries to items whose part was already destroyed, a share that only grows over a session.

Same scenarios and install: unfixed build 1 exception, patched build 0. A wider pass of 77 editor cases is clean.

### DIFF
--- a/KSPCommunityFixes/QoL/UIFloatEditNumericInput.cs
+++ b/KSPCommunityFixes/QoL/UIFloatEditNumericInput.cs
@@ -120,12 +120,30 @@ namespace KSPCommunityFixes
             fieldNameNumeric.text = field.guiName;
             inputField.onEndEdit.AddListener(OnFieldInput);
             inputField.onSelect.AddListener(AddInputFieldLock);
-            GameEvents.onPartActionNumericSlider.Add(ToggleNumericSlider);
+            // OnEnable() has already run by now (the window instantiates the item and activates it
+            // before calling Setup()) and couldn't subscribe because the item had no field yet, so
+            // do it here. Any further enable/disable cycle is handled by OnEnable()/OnDisable().
+            if (isActiveAndEnabled)
+                GameEvents.onPartActionNumericSlider.Add(ToggleNumericSlider);
             ToggleNumericSlider(GameSettings.PAW_NUMERIC_SLIDERS);
             window.usingNumericValue = true;
         }
 
-        private void OnDestroy()
+        private void OnEnable()
+        {
+            // onPartActionNumericSlider is a global event, so only stay subscribed while this item
+            // is actually shown. An item that isn't (a collapsed PAW group, a closed window, or the
+            // prefab itself) would otherwise keep handling toggles, and would push its input field
+            // content into a part or module that may not even exist anymore.
+            if (field == null)
+                return;
+
+            GameEvents.onPartActionNumericSlider.Add(ToggleNumericSlider);
+            // toggles that happened while we were disabled were missed, so resync
+            ToggleNumericSlider(GameSettings.PAW_NUMERIC_SLIDERS);
+        }
+
+        private void OnDisable()
         {
             GameEvents.onPartActionNumericSlider.Remove(ToggleNumericSlider);
         }
@@ -146,7 +164,14 @@ namespace KSPCommunityFixes
             }
             else
             {
-                OnFieldInput(inputField.text);
+                // Only push the input field content back to the field if it was actually edited.
+                // Doing this unconditionally results in a ton of unnecessary change callbacks
+                // and rounds the field value unnecessarily.
+                if (inputField.text != GetFieldValue().ToString($"F{floatControl.sigFigs}"))
+                    OnFieldInput(inputField.text);
+                else
+                    RemoveInputfieldLock();
+
                 float value = GetFieldValue();
                 string unit = floatControl.unit;
                 int sigFigs = floatControl.sigFigs;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 **Bug Fixes**
 - **EditorAnimatedPartsShipModified** : Fixed a memory leak ([issue #396](https://github.com/KSPModdingLibs/KSPCommunityFixes/issues/396)) where listeners were not properly cleaned up.
 - **EditorAnimatedPartsShipModified** : This patch is now disabled if DMagic Orbital Science is installed because DMagic Orbital Science has a bug that causes a stack overflow crash if you ever call `DMSoilMoisture.OnStop`, which this patch does.
+- **UIFloatEditNumericInput** : Turning the numeric input ("#") mode off no longer re-applies the input field value to PAW items whose part or module has been destroyed, which could throw from the field change callbacks of third party modules. The value is also only re-applied when the input field was actually edited, instead of on every item on every toggle, which was silently rounding every float edit field value to its displayed precision.
 
 ##### 1.41.1
 **Bug Fixes**


KSPCommunityFixes/QoL/UIFloatEditNumericInput.cs
```diff
@@ -146,7 +160,17 @@ private void ToggleNumericSlider(bool numeric)
...
+                // Only push the input field content back to the field if it was actually edited.
+                // Doing this unconditionally means that switching the numeric mode off re-applies
+                // the value of every numeric field of every part having a PAW item, firing all the
+                // associated change callbacks. On top of being wasteful, the re-applied value is
+                // the rounded one that was shown in the input field, so this silently degrades the
+                // precision of every such field on every numeric mode toggle.
+                if (inputField.text != GetFieldValue().ToString($"F{floatControl.sigFigs}"))
+                    OnFieldInput(inputField.text);
+                else
+                    RemoveInputfieldLock();
```
Phantomical, line 173: This seems reasonable to me. Could you compress down the comment a bit though?
```suggestion
                // Only push the input field content back to the field if it was actually edited.
                // Doing this unconditionally results in a ton of unnecessary change callbacks
                // and rounds the field value unnecessarily.
                if (inputField.text != GetFieldValue().ToString($"F{floatControl.sigFigs}"))
                    OnFieldInput(inputField.text);
                else
                    RemoveInputfieldLock();
```

KSPCommunityFixes/QoL/UIFloatEditNumericInput.cs
```diff
@@ -132,6 +132,20 @@ private void OnDestroy()
...
+            // and subscribed, as we only unsubscribe from OnDestroy(), which Unity won't call at
+            // all on a component that was never activated, and which is deferred by at least a
+            // frame otherwise. Going further would push the input field value into a dead part,
+            // firing the module change callbacks and usually resulting in an exception being
+            // thrown from the module code.
+            if (this.IsDestroyed() || part.IsNullOrDestroyed() || (isModule && partModule.IsNullOrDestroyed()))
+            {
+                GameEvents.onPartActionNumericSlider.Remove(ToggleNumericSlider);
+                return;
+            }
```
Phantomical: I don't think this is the right way to go about this. Could you instead change the code here to add the event listener in `OnEnable` and remove it in `OnDisable`? That way we are actually removing the listeners proactively instead of only when an event happens.

sparr:
The suggestion on the write-back comment is in verbatim.

For the listener: the destroyed-object guard is gone, `Setup()` no longer subscribes
unconditionally, and `OnDestroy()` is gone, all replaced by `OnEnable()`/`OnDisable()`

Three things that came out of doing it this way, in case you want them done differently:

`Setup()` still subscribes, guarded by `isActiveAndEnabled`. `UIPartActionWindow`
instantiates the item, calls `SetActive(true)`, and only then calls `Setup()`, so
`OnEnable()` always runs first, at a point where `field` is still null and there is
nothing to subscribe on behalf of. The `isActiveAndEnabled` check is what stops the
two paths from double-subscribing if an item is ever set up while inactive.

The `field == null` return also covers the prefab: `CreatePrefab()` does
`AddComponent<UIPartActionNumericFloatEdit>()`, which fires `OnEnable()` right away if
that GameObject is active, and the prefab never gets a `Setup()` call.

`OnEnable()` re-applies `GameSettings.PAW_NUMERIC_SLIDERS`. Previously a hidden item
(collapsed group, closed window) stayed subscribed and tracked toggles it couldn't
display; now it doesn't get them, so without the resync it can come back in the wrong
mode. Happy to drop that line if you'd rather keep the change minimal.

One case the old guard covered that this doesn't: an item still active in the hierarchy
whose part was destroyed earlier in the same frame, since Unity flags the object
destroyed immediately but defers `OnDisable` to end of frame. The crash I actually hit
isn't that case, though - the item there had `activeInHierarchy=False`, so `OnEnable`/
`OnDisable` handles it.

